PR #1, opened 2022-11-29
Consolidate IPNI spec from docs and other repos

masih:
Gather specification of IPNI protocol from `docs.cid.contact`, and `storetheindex` docs into a single specification.

[Rendered](https://github.com/ipni/specs/blob/d28e864fba9ac7e7d064b512636e21a386a12993/IPNI.md).


IPNI.md
```diff
@@ -0,0 +1,633 @@
+# IPNI: InterPlanetary Network Indexer
+
+![reliable](https://img.shields.io/badge/status-reliable-green.svg?style=flat-square)
+
+**Author(s)**:
```
masih, line 5: Please let me know if there is anyone I have missed. Thanks

PR #5, opened 2022-12-09
Add reader's privacy specification

ischasny:
This PR introduces IPNI reader's privacy specification. This is a part of larger IPFS privacy programme TODO: paste link here once it's done.


reader-privacy.md
```diff
@@ -0,0 +1,123 @@
...
+* In response to the hashed find request, the indexer will return a set of encrypted `ProviderRecordKey`s. 
+`ProviderRecordKey` will consist of two concatenated hashes - one over `peerID` and the other over `contextID`. 
+Each `ProviderRecordKey` will be encrypted with a key derived from the *original* CID value: 
+`enc(hash(peerID) || hash(contextID), CID)`, where `hash` is a hash over the value, `||` is concatenation 
+and `enc` is encryption over the value. In order to make sense of that payload, a passive observer would need 
+to get hold of the original CID that isn't revealed during the communication round;
+* Using the original CID, the client would decrypt `ProviderRecordKey`s and then calculate another hash
+over the decrypted `hash(peerID)` part of it. Using that hash for each `ProviderRecordKey` the client would do another lookup 
+to get an encrypted `ProviderRecord` in response. `ProviderRecord` will contain information about provider, 
+such as it's *peerID*, *multiaddresses*, *supported protocols* and etc. Each `ProviderRecord` will be encrypted
```
masih, line 92:
```suggestion
such as it's *peerID*, *multiaddrs*, *supported protocols* and so on. Each `ProviderRecord` will be encrypted 
```

reader-privacy.md
```diff
@@ -0,0 +1,123 @@
...
+* Using the original CID, the client would decrypt `ProviderRecordKey`s and then calculate another hash
+over the decrypted `hash(peerID)` part of it. Using that hash for each `ProviderRecordKey` the client would do another lookup 
+to get an encrypted `ProviderRecord` in response. `ProviderRecord` will contain information about provider, 
+such as it's *peerID*, *multiaddresses*, *supported protocols* and etc. Each `ProviderRecord` will be encrypted 
+with a key derived from `hash(peerID)`. In order to make sense of that payload, a passive observer would need to 
+get hold of the decrypted `ProviderRecordKey` that isn't revealed during the communication round;
+* Using the `hash(peerID)` from `ProviderRecordKey`s, the client would decrypt `ProviderRecord`s and then reach out to the 
+provider directly to fetch the desired content. 
+
+By utilising such scheme only a party that knows original CID that is being looked up can decode the protocol,
```
masih, line 98:
```suggestion
By utilising such scheme only a party that knows original CID can decode the protocol,
```

reader-privacy.md
```diff
@@ -0,0 +1,123 @@
...
+however that CID is never revealed.
```
masih, line 99:
```suggestion
and that CID is never revealed. 
```

reader-privacy.md
```diff
@@ -0,0 +1,123 @@
...
+get hold of the decrypted `ProviderRecordKey` that isn't revealed during the communication round;
+* Using the `hash(peerID)` from `ProviderRecordKey`s, the client would decrypt `ProviderRecord`s and then reach out to the 
+provider directly to fetch the desired content. 
+
+By utilising such scheme only a party that knows original CID that is being looked up can decode the protocol,
+however that CID is never revealed. 
+
+### Security
+​
+Security model of the Reader's Privacy proposal boils down to inability to *algorithmically* derive the original CID value for a
```
masih, line 103: I would drop `'s` so that the spec rolls off the tongue easier.

```suggestion
Security model of the Reader Privacy proposal boils down to inability to *algorithmically* derive the original CID value for a 
```

reader-privacy.md
```diff
@@ -0,0 +1,123 @@
...
+* Using the `hash(peerID)` from `ProviderRecordKey`s, the client would decrypt `ProviderRecord`s and then reach out to the 
+provider directly to fetch the desired content. 
+
+By utilising such scheme only a party that knows original CID that is being looked up can decode the protocol,
+however that CID is never revealed. 
+
+### Security
+​
+Security model of the Reader's Privacy proposal boils down to inability to *algorithmically* derive the original CID value for a 
+`hash(CID)` that is used for IPNI lookups. Right now indexer advertisments are not encrypted, but authenticated and contain plain CID values in them.
```
masih, line 104: Early on we used the phrase "indexer" a lot. We even had "indexer provider" which is not a nice name. 

We should avoid it if we can and simply refer to advertisements as yes you guessed it: "advertisements". If you have to qualify that phrase then my pick would be to use "IPNI advertisements".

```suggestion
`hash(CID)` that is used for IPNI lookups. Right now advertisments are not encrypted, but authenticated and contain plain CID values in them. 
```

reader-privacy.md
```diff
@@ -0,0 +1,123 @@
...
+provider directly to fetch the desired content. 
+
+By utilising such scheme only a party that knows original CID that is being looked up can decode the protocol,
+however that CID is never revealed. 
+
+### Security
+​
+Security model of the Reader's Privacy proposal boils down to inability to *algorithmically* derive the original CID value for a 
+`hash(CID)` that is used for IPNI lookups. Right now indexer advertisments are not encrypted, but authenticated and contain plain CID values in them. 
+That is going to change once *Writer's Privacy* is implemented. Before that a sophisticated attacker could build a map of `hash(CID) -> CID`
```
masih, line 105: It doesn't take that much sophistication.

```suggestion
That is going to change once *Writer Privacy* is implemented. Until then, an attacker could build a map of `hash(CID) -> CID` 
```

reader-privacy.md
```diff
@@ -0,0 +1,123 @@
...
+By utilising such scheme only a party that knows original CID that is being looked up can decode the protocol,
+however that CID is never revealed. 
+
+### Security
+​
+Security model of the Reader's Privacy proposal boils down to inability to *algorithmically* derive the original CID value for a 
+`hash(CID)` that is used for IPNI lookups. Right now indexer advertisments are not encrypted, but authenticated and contain plain CID values in them. 
+That is going to change once *Writer's Privacy* is implemented. Before that a sophisticated attacker could build a map of `hash(CID) -> CID` 
+by re-ingesting advertisements chain from each publisher so that they can use it to decrypt the protocol. 
+Doing that will require significant investment into infrastructure and will be eliminated as a possibility after *Writer's Privacy* upgrade.
```
masih, line 107:
```suggestion
Doing that will require significant resources as it involves crawling the entire network. However, it will eventually be eliminated by *Writer Privacy* upgrade.
```

reader-privacy.md
```diff
@@ -0,0 +1,123 @@
...
+
+By utilising such scheme only a party that knows original CID that is being looked up can decode the protocol,
+however that CID is never revealed. 
+
+### Security
+​
+Security model of the Reader's Privacy proposal boils down to inability to *algorithmically* derive the original CID value for a 
+`hash(CID)` that is used for IPNI lookups. Right now indexer advertisments are not encrypted, but authenticated and contain plain CID values in them. 
+That is going to change once *Writer's Privacy* is implemented. Before that a sophisticated attacker could build a map of `hash(CID) -> CID` 
+by re-ingesting advertisements chain from each publisher so that they can use it to decrypt the protocol.
```
masih, line 106:
```suggestion
by re-ingesting advertisements chain from each publisher in order to collect all original CIDs which can then be used to decrypt provider records and so on. 
```

reader-privacy.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Security
+​
+Security model of the Reader's Privacy proposal boils down to inability to *algorithmically* derive the original CID value for a 
+`hash(CID)` that is used for IPNI lookups. Right now indexer advertisments are not encrypted, but authenticated and contain plain CID values in them. 
+That is going to change once *Writer's Privacy* is implemented. Before that a sophisticated attacker could build a map of `hash(CID) -> CID` 
+by re-ingesting advertisements chain from each publisher so that they can use it to decrypt the protocol. 
+Doing that will require significant investment into infrastructure and will be eliminated as a possibility after *Writer's Privacy* upgrade.
+
+Reader's Privacy is a first step towards fully private content routing protocol.
```
masih, line 109:
```suggestion
Reader Privacy is a first step towards fully private content routing protocol. 
```

reader-privacy.md
```diff
@@ -0,0 +1,123 @@
...
+​
+Security model of the Reader's Privacy proposal boils down to inability to *algorithmically* derive the original CID value for a 
+`hash(CID)` that is used for IPNI lookups. Right now indexer advertisments are not encrypted, but authenticated and contain plain CID values in them. 
+That is going to change once *Writer's Privacy* is implemented. Before that a sophisticated attacker could build a map of `hash(CID) -> CID` 
+by re-ingesting advertisements chain from each publisher so that they can use it to decrypt the protocol. 
+Doing that will require significant investment into infrastructure and will be eliminated as a possibility after *Writer's Privacy* upgrade.
+
+Reader's Privacy is a first step towards fully private content routing protocol. 
+
+Wider security implications are discussed in the IPFS Reader's Privacy specification: TODO link here.
```
masih, line 111:
```suggestion
Wider security implications are discussed in the IPFS Reader Privacy specification: TODO link here.
```

masih (commented):
Thank you for adding this @ischasny 👍 

Github markdown supports [mermaid](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/) diagrams. If you are up to it, the specification would fantastically benefit from a flowchart and make it a lot easer for the reader to wrap their head around interactions.

I recommend adding a section that adds some colour in terms of cost implications: for the indexers in terms of storage and for the client in terms of extra hops.

reader-privacy.md
```diff
@@ -0,0 +1,123 @@
...
+![Index building flow](resources/readers-privacy-1.png)
+
+## Specification
+​
+This specification focuses on improving the **step #3** where a client has to pass a CID to the indexer *in open* 
+to get a list of providers where the content can be fetched from. 
+
+In order to protect the reader's privacy the proposal is to change the way how CID lookup works to the following:
+
+* A client who wants to do a lookup will calculate a hash over the CID (`hash(CID)`) and use it for the
```
aschmahmann, line 81: This is `hash(multihash component of cid)` right?

reader-privacy.md
```diff
@@ -0,0 +1,123 @@
...
+lookup query (hence the name double hashing);
+* In response to the hashed find request, the indexer will return a set of encrypted `ProviderRecordKey`s. 
+`ProviderRecordKey` will consist of two concatenated hashes - one over `peerID` and the other over `contextID`. 
+Each `ProviderRecordKey` will be encrypted with a key derived from the *original* CID value:
```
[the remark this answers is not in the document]
ischasny: It was about a nonce derived from context id unless I'm missing something?

reader-privacy.md
```diff
@@ -0,0 +1,120 @@
...
+This is obviously undesirable and has been for some time now a strong request from the community.
+
+The changes described in this specification introduce a IPNI Readres Privacy upgrade. It will prevent 
+passive observers from tracking user's actions as described above. It will also be a first step towards
+fully private IPNI protocol that will eliminate indexers as centralised observers.  
+
+### Non Goals
+
+* Writer's (publisher's) Privacy, which is going to be done as a separate specification;
+* Client to Provider privacy, that is out of scope for the content routing subsystem.
```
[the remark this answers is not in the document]
ischasny: Yes, that sounds better - applying

reader-privacy.md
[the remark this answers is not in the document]
masih: I am hallucinating again thank you

ischasny:
Thanks @masih , great suggestion on adding a mermaid diagram - done.

reader-privacy.md
```diff
@@ -0,0 +1,142 @@
...
+In order to protect the reader's privacy the proposal changes the way CID lookup works to the following:
+
+* A client who wants to do a lookup will calculate a hash over the CID's multihash (`hash(MH)`) and use it for the
+lookup query (hence the name double hashing);
+* In response to the hashed find request, the indexer will return a set of encrypted `ProviderRecordKey`s. 
+`ProviderRecordKey` will consist of two concatenated hashes - one over `peerID` and the other over `contextID`. 
+Each `ProviderRecordKey` will be encrypted with a key derived from the *original* multihash value: 
+`enc(hash(peerID) || hash(contextID), MH)`, where `hash` is a hash over the value, and `||` is concatenation 
+and `enc` is encryption over the value. In order to make sense of that payload, a passive observer would need 
+to get hold of the original CID that isn't revealed during the communication round;
```
willscott, line 88: Instead of `enc(hash(peerID) || hash(contextID), MH)` can we use`enc(peerID || contextID, MH)` ?

then the client decodes `peerID` and calculates `hash(peerID)` for it's subsequent query

ischasny: The problem with full peer IDs in datastore keys (besides extra space required) is that they will make value keys be variable length. All current datastores rely on keys being fixed length so that they can be efficiently stored in arrays. 

Is your concern hash collisions?

willscott: my concerns are:
1. Getting back the original peerID from the query means that this can continue to fullfill the current ContentRouting API in IPFS
2. Compatibility with the form that double-hashing for DHT is taking.

At one point masih proposed an additional level of 'pointer' indirection: in primary values table we point to another table of these `enc(ProviderRecordKey)` values - such that the record keys can be of variable length. (when used as keys, the key form is still `hash(ProviderRecordKey)`, so will remain fixed length)

In practice these should be almost all the same length - we could potentially partition the value table based on `len(ProviderRecordKey)` and have a few different value tables where we search each of them? (i imagine we'd have one for when a deal id is specified, and one for entries without a contextID))

ischasny: Added to the latest revision of the spec.

reader-privacy.md
```diff
@@ -0,0 +1,199 @@
...
+It can be reduced down to one by caching `ProviderRecord`s locally at the client side. That would eliminate a need in a lookup
+per decrypted `ProviderRecordKey`. In the future there can be a separate service that distributes `PeerID` to `Multiaddresses` mappings in open.
+That dataset can be periodically downloaded by all clients and cached locally;
+
+* **Extra compute**. At minimum, clients will require to perform an extra hashing per CID and decryption per `ProviderRecordKey` that will add
+some overhead to each lookup. That overhead can be initially offloaded to the server but will have to be done by clients eventually (explained below);
+
+* **Extra storage space**. Storing encrypted data will require more space due to paddings and nonce;
+
+* **Bulk deletes**. Encrypted `PeerID` will be different for each multihash and hence bulk delete operations (delete everything for a provider X) will not be possible.
```
guillaumemichel, line 160: > Encrypted `PeerID`

What is the purpose of encrypting the `PeerID`?

reader-privacy.md
```diff
@@ -0,0 +1,199 @@
...
+by re-ingesting advertisements chain from each publisher in order to collect all original multihashes which can then be used to decrypt provider records and so on. 
+Doing that will require significant resources as it involves crawling the entire network. However, it will eventually be eliminated by *Writer Privacy* upgrade.
+
+Reader Privacy is a first step towards fully private content routing protocol. 
+
+Wider security implications are discussed in the IPFS Reader Privacy specification: TODO link here.
+
+#### Hashing and Encryption Function Upgrades
+
+All multihashes have a codec encoded in them. If a hashing or encryption funciton will have to rotate then different types of multihahses can coexist together
```
guillaumemichel, line 137:
```suggestion
All multihashes have a codec encoded in them. If a hashing or encryption function will have to rotate then different types of multihashes can coexist together
```

reader-privacy.md
```diff
@@ -0,0 +1,199 @@
...
+and can be processed differently by IPNI implementations. It won't be possible to apply a fix retroactivelly to the data returned by previous lookup requests, 
+however IPNI implementations should start blocking all new ones that use a compromised scheme.
```
guillaumemichel, line 139: so IPNI will not allow retro compatibility with older client versions as soon as a fix is issued? Clients will need to upgrade to keep using the IPNI?

reader-privacy.md
```diff
@@ -0,0 +1,199 @@
...
+
+* **Operational overhead**. Reader Privacy roll out will be a gradual process as many clients will have to migrate over. During the transition period 
+IPNI implementations will have to serve both plain and hashed lookups. That will involve either:   
+    - spinning up separate IPNI instances for hashed queries or;
+    - serving hashed and regular queries from the same instances using encrypted dataset. That means that servers will have to do decryption on behalf of their clients
+    using a plain multihash that has been provided in the lookup request;
+
+* **Data Migration**. Existing indexes will have to undergo data migration or to be wiped out complletely and rebuilt again. 
+
+### Threat Modelling
```
guillaumemichel, line 170: Is there anything preventing content providers to publish countless dummy provider records to exhaust the IPNI?

ischasny: Great question. The protocol itself doesn't enforce any restrictions on how many records can be published in a period of time. It's up to the implementation to implement rate limiting and block abusive storage providers.

reader-privacy.md
ischasny: Great spot! There should be some transition period of course. Rephrased to make it clearer.

reader-privacy.md
ischasny: On ingestion each multihash is going to be hashed and each peerID is going to be encrypted with a key derived from the multihash value. IPNI will store `hash(multihash) -> enc(peerID, multihash)` records so that it can return an encrypted peerID to a hashed lookup request. Otherwise it will have to return a raw peerID value that would leak identity of the provider to a passive observer.

reader-privacy.md
```diff
@@ -0,0 +1,199 @@
...
+index can be migrated over to new functions by reingesting all advertisement chains. With Writer Privacy, Publishers will have to republish advertisments 
+using new algorithms. Both old and new scheme can coexist together for some time. The old one should be retired either immediately or once
+the indexes have been rebuilt and the users have been migrated over. 
+
+Exact operation procedure will be different for differnet IPNI implementations.
+
+### Trade Offs 
+
+* **Multiple lookups**. In the simplest scenario Reader Privacy protocol will require at least two roundtrips to find provider details for a given CID.
+It can be reduced down to one by caching `ProviderRecord`s locally at the client side. That would eliminate a need in a lookup
```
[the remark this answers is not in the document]
ischasny: Yes, exactly. This is what `ProviderRecord` is - peer id combined with multiaddresses and some other metadata (like supported protocols)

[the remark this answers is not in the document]
guillaumemichel: Alright I got confused because we use the name `ProviderRecord` for the `CID -> PeerID` mapping (or `Hash(mulithash) -> PeerID`) and the name `PeerRecord` for the `PeerID -> Multiaddr` mapping.

The `ProviderRecordKey` and `ProviderRecord` are well defined in the _Specification_ section, I must have missed it while reading 👍🏻

reader-privacy.md
```diff
@@ -0,0 +1,199 @@
...
+* In response to the hashed find request, the indexer will return a set of encrypted `ProviderRecordKey`s. 
+`ProviderRecordKey` will consist of the `peerID` concatenated with a hash over `contextID`. 
+Each `ProviderRecordKey` will be encrypted with a key derived from the *original* multihash value: 
+`enc(peerID || hash(contextID), MH)`, where `hash` is a hash over the value, and `||` is concatenation 
+and `enc` is encryption over the value. In order to make sense of that payload, a passive observer would need 
+to get hold of the original CID that isn't revealed during the communication round;
+* Using the original multihash, the client will decrypt `ProviderRecordKey`s and then calculate a hash
+over the decrypted `peerID` part of it. Using such hash for each `ProviderRecordKey` the client would do another lookup 
+to get an encrypted `ProviderRecord` in response. `ProviderRecord` will contain information about the provider, 
+such as it's *peerID*, *multiaddrs*, *supported protocols* and so on. Each `ProviderRecord` will be encrypted
```
masih, line 95: How does the lookup for `ProviderRecord` work if the lookup key consists of the hash of `peerID` only? `ProviderRecord` is uniquely identified by `PeerID` _and_ `ContextID`; right?

Is the idea that the server will send back the list of _all_ `ProviderRecord`s for the given key and the client would filter by `ContextID` on their side to avoid revealing the `ContextID` to the server?

ischasny: Great question. By `ProviderRecord` here I mean a tuple of `peerID` to a list of `multiaddresses` (i.e. without the deal-specific metadata). I will additionally clarify this in the spec.

masih: Thank you; what about context ID? 

It looks like the returned `hash(contextID)` is not used at all. How would the client retrieve the IPNI `Metadata`?

reader-privacy.md
```diff
@@ -0,0 +1,199 @@
...
+* In response to the hashed find request, the indexer will return a set of encrypted `ProviderRecordKey`s. 
+`ProviderRecordKey` will consist of the `peerID` concatenated with a hash over `contextID`. 
+Each `ProviderRecordKey` will be encrypted with a key derived from the *original* multihash value: 
+`enc(peerID || hash(contextID), MH)`, where `hash` is a hash over the value, and `||` is concatenation 
+and `enc` is encryption over the value. In order to make sense of that payload, a passive observer would need 
+to get hold of the original CID that isn't revealed during the communication round;
+* Using the original multihash, the client will decrypt `ProviderRecordKey`s and then calculate a hash
+over the decrypted `peerID` part of it. Using such hash for each `ProviderRecordKey` the client would do another lookup 
+to get an encrypted `ProviderRecord` in response. `ProviderRecord` will contain information about the provider, 
+such as it's *peerID* and *multiaddrs*. Each `ProviderRecord` will be encrypted
```
masih, line 95:
```suggestion
consisting of its *peerID* and *multiaddrs*. Each `ProviderRecord` will be encrypted 
```

reader-privacy.md
ischasny: The idea is to leave it for future improvements. In next iteration we can make `hash(contextID)` to be an optional input parameter that would add context-specific data to the returned payload.

reader-privacy-addendum.md
```diff
@@ -0,0 +1,30 @@
+# Reader Privacy Implementation Details
```
masih, line 1: Can we move this to a single doc, under `Implementation/AESGSM`or `Implementers Considerations/AESGSM` section please?

ischasny: Done

reader-privacy.md
```diff
@@ -0,0 +1,242 @@
...
+
+### Hashing
+
+SHA256 must be used as a hashing function. To avoid collisions between the hashes calculated from the same value but for different purposes (for example
+to derive an encryption key versus derive a lookup key) a constant string can be appended to the value before calculating a hash over it. For example
+`sha256("AESGCM" + multihash)` to derive an encryption key and `sha256("CR_DOUBLEHASH" + multihash)` to calculate a second hash over the multihash.
+
+All hashed data that is used for lookups must be of `Multihash` format with `SHA_256` codec. Double hashed data must use `DBL_SHA_256` codec.
+
+Multihashes must be prepended with `CR_DOUBLEHASH` before calculating a second hash. Unhashed data must be prepended with `CR_HASH` before calculating the first hash.
```
aschmahmann, line 213: Are you sure you're meaning to use the DBL_SHA_256 code? It looks like that code and implementation are already defined. e.g. https://github.com/multiformats/go-multihash/blob/608669da49b636a646de3472101d0183889ae6c4/core/errata.go#L29.

`SHA256("CR_DOUBLEHASH" || SHA256(CR_HASH, <data>) != SHA256(SHA256(<data>))` which are both different from `SHA256("CR_DOUBLEHASH" || SHA256-MH(CR_HASH, <data>)`

reader-privacy.md
```diff
@@ -0,0 +1,242 @@
...
+over the decrypted `peerID` part of it. Using such hash for each `ProviderRecordKey` the client would do another lookup 
+to get an encrypted `ProviderRecord` in response. `ProviderRecord` will contain information about the provider, 
+consisting of its *peerID* and *multiaddrs*. Each `ProviderRecord` will be encrypted 
+with a key derived from `peerID`. In order to make sense of that payload, a passive observer would need to 
+get hold of the original `peerID` that isn't revealed during the communication round;
+* Using a key derived from `peerID`, the client will decrypt `ProviderRecord`s and then reach out to the 
+provider directly to fetch the desired content. 
+* The client might choose to fetch additional `Metadata` that is supplied to IPNI in Advertisements.
+That will require another lookup by `hash(ProviderRecordKey)` to get `enc(Metadata, ProviderRecordKey)` in response. 
+*This step will not be required for IPFS as Bitswap protocol is assumed implicitly.*
```
aschmahmann, line 108: I'm confused by what you mean here (aside from the general IPFS == Bitswap thing which is likely to confuse others).

Do you mean that implementations that rely on libp2p protocol negotiation with multistream can choose to ignore the metadata request because they can just talk to the target peer directly (e.g. true for Bitswap, vanilla GraphSync, etc.)?

reader-privacy.md
```diff
@@ -0,0 +1,242 @@
...
+* A client who wants to do a lookup will calculate a hash over the CID's multihash (`hash(MH)`) and use it for the
+lookup query (hence the name double hashing);
+* In response to the hashed find request, the indexer will return a set of encrypted `ProviderRecordKey`s. 
+`ProviderRecordKey` will consist of the `peerID` concatenated with `contextID`. 
+Each `ProviderRecordKey` will be encrypted with a key derived from the *original* multihash value: 
+`enc(peerID || contextID, MH)`, where `hash` is a hash over the value, and `||` is concatenation 
+and `enc` is encryption over the value. In order to make sense of that payload, a passive observer would need 
+to get hold of the original CID that isn't revealed during the communication round;
+* Using the original multihash, the client will decrypt `ProviderRecordKey`s and then calculate a hash
+over the decrypted `peerID` part of it. Using such hash for each `ProviderRecordKey` the client would do another lookup
```
aschmahmann, line 99: This seems to me just as optional as fetching `Metadata`. In particular, given the cost of an extra round-trip I could see implementations skipping the multiaddr lookup if they already know the addresses from a previous lookup

reader-privacy.md
```diff
@@ -0,0 +1,242 @@
...
+* In response to the hashed find request, the indexer will return a set of encrypted `ProviderRecordKey`s. 
+`ProviderRecordKey` will consist of the `peerID` concatenated with `contextID`. 
+Each `ProviderRecordKey` will be encrypted with a key derived from the *original* multihash value: 
+`enc(peerID || contextID, MH)`, where `hash` is a hash over the value, and `||` is concatenation 
+and `enc` is encryption over the value. In order to make sense of that payload, a passive observer would need 
+to get hold of the original CID that isn't revealed during the communication round;
+* Using the original multihash, the client will decrypt `ProviderRecordKey`s and then calculate a hash
+over the decrypted `peerID` part of it. Using such hash for each `ProviderRecordKey` the client would do another lookup 
+to get an encrypted `ProviderRecord` in response. `ProviderRecord` will contain information about the provider, 
+consisting of its *peerID* and *multiaddrs*. Each `ProviderRecord` will be encrypted
```
aschmahmann, line 101: If I'm reading this correctly this is sending back `enc(peerID || multiaddrs, peerID)` why are you sending back the peerID when the client already has it?

reader-privacy.md
```diff
@@ -0,0 +1,242 @@
...
+on the diagram below:
+​
+![Index building flow](resources/readers-privacy-1.png)
+
+## Specification
+​
+This specification improves the reader privacy by proposing changes to the Step 3, depicted above, where the client 
+supplies the content CID directly in order to lookup its corresponding providers.
+
+In order to protect the reader's privacy the proposal changes the way CID lookup works to the following:
```
aschmahmann, line 88: Perhaps I'm missing something obvious, why does this require making 3 round-trips instead of one? I suspect this is related to some plans you have in mind for Writer Privacy, but at first glance it seems like the server should be able to respond with all the information (peerID, multiaddrs, metadata) in a single response encrypted using the multihash.

reader-privacy.md
ischasny: Fair point, removed the `peerID`

reader-privacy.md
ischasny: Good point - removed that sentence. Let me explain myself.  

The client must be able to understand what protocol to use to speak to the returned providers. As all data on the IPNI server is encrypted - including metadata that defines the protocol - the server can not apply any filtering logic, as that would require knowledge of the original multihash which the client doesn't want to reveal. 

Having said that - filtering logic would have to be applied on the client side either by querying metadata or by doing libp2p negotiation.

reader-privacy.md
ischasny: Added a sentence about `AddrInfo` being cacheable.

reader-privacy.md
ischasny: This is to be aligned with the current [DHT implementation](https://github.com/ChainSafe/go-libp2p-kad-dht/blob/dc84e3ec9de47811c1d28d5b644cc34e4a6c4d53/internal/utils.go#L12). It uses `DBL_SHA_256` codec as well as prepends the bytes with `CR_DOUBLEHASH` before calculating a second multihash.

guillaumemichel: @aschmahmann do you suggest defining a new constant string for the purpose of double hashing in both the DHT and Indexers?

reader-privacy.md
```diff
@@ -0,0 +1,242 @@
...
+​
+Security model of the Reader Privacy proposal boils down to inability of an attacker to *algorithmically* derive the original multihash value from 
+`hash(multihash)` that is used for IPNI lookups. IPNI advertisments are not encrypted, but authenticated and contain plain multihash values in them.
+Before Writer Privacy is implemented an attacker could build a map of `hash(multihash) -> multihash` 
+by re-ingesting advertisements chain from each publisher in order to collect all original multihashes which can then be used to decrypt provider records and so on. 
+Doing that will require significant resources as it involves crawling the entire network. However, it will eventually be eliminated by *Writer Privacy* upgrade.
+
+Reader Privacy is a first step towards fully private content routing protocol. 
+
+Wider security implications are discussed in the IPFS Reader Privacy specification: TODO link here.
```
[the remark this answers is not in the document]
ischasny: Thanks @aschmahmann , great points, included into the specification.

BigLep:
Hi IPNI team.  No pressure/rush on this, but I wanted to confirm that that will be an IPIP at some point to ipfs/specs for changes https://github.com/ipfs/specs/blob/main/routing/DELEGATED_CONTENT_ROUTING_HTTP.md and then a corresponding go-libipfs change to https://github.com/ipfs/go-libipfs/tree/main/routing/http

I'm asking to make sure I'm thinking about this right and planning for it accordingly.  @Jorropo will be the POC to engage with from the EngRes IPFS Stewards side.

ischasny:
> https://github.com/ipfs/go-libipfs/tree/main/routing/http

Hi @BigLep  - yes, that's the plan. Thanks for pointing that out.

gammazero (approved):

masih (approved):

PR #7, opened 2023-01-18
Specify semantics of empty metadata in advertisements

masih:
Update specs to spell out behaviour of indexer when metadata is empty.


IPNI.md
```diff
@@ -293,6 +293,8 @@ Specified protocols are expected to be ordered in increasing order.
     * the proposed `uvarint` protocol is `0x3D0000`.
     * the following bytes are not yet defined.
 
+If the `Metadata` field is not specified, the advertisement is treated as address update only.
```
ischasny, line 296: Shall it also say about empty addresses? That those will be skipped.

Issue #8, opened 2023-01-26
Refine `GET /multihash/<multihash>` response for `ndjson`


masih:
The response structure used by IPNI for `GET /multihash/<multihash>` is optimised for batch find: it includes the multihash that is being looked up as base64 encoded string, along with a nested array of results. 

I believe the rationale has been to use the same response model for both batch find, and single find request. For non-streaming responses this is fine; however, for responses streamted using new-line delimited json (i.e. ndjson) it becomes suboptimal:
* in ndjson each response needs to be a JSON object.
* because the results for a single mulithash are nested in an array, to stream the response using the current model one has to repeat the multihash that is being looked up in every streamed response.

This is less than optimum considering the vast majority of traffic received by cid.contact is for explicit multihash lookups.

Consider spec changes for a more optimal response structure for single multihash lookup, such that:
* the original multihash looked up via `GET` request is not repeated in the response or optionally included.
* the response model is preferably made up of a series of self contained JSON objects optimised for the streaming case, which hopefully becomes the default/preferred response media type when `Accept` request header permits it.

PR #9, opened 2023-01-26
Introduce specification for cascading lookup query parameter

masih:
It is increasingly useful for the users to be able to search alternative routing systems through the familiar IPNI endpoints. This greatly simplifies the barrier to simply find providers for a given CID, and avoid the need to learn about explicit routing systems such as the DHT. Further, the results can be cached and reused by others considering the fact that a typical IPNI instance will heavily utilise caching mechanisms regardless of cascading lookups.

The cascading lookup must be configurable to avoid doubling load on other routing systems in a case where, having searched other routing systems already, the user exclusively wants to search IPNI. For example, Kubo `v0.18.0` is configured by default to lookup records on the DHT and on IPNI. Therefore, automatic cascading of lookups onto the DHT by IPNI would result in duplication of load and waisted effort since Kubo independently looks up the DHT.

The changes here introduce a new query parameter, called `cascade` for the lookup APIs, where the user can specify a list of comma separated alternative routing systems to which the query is cascaded in addition to IPNI. By default no cascading occurs. This allows Kubo to gracefully continue functioning while enabling other clients to optionally cascade lookups onto the IPFS DHT.

### DIFF
--- a/IPNI.md
+++ b/IPNI.md
@@ -469,6 +469,11 @@ and uses the multihash portion of the CID only.
 * `cid` - _Required_. The default string representation of the Cid. Currently, Base32 is used for
   CIDv1 as the encoding for the multibase string and Base58 is used for CIDv0.
 
+##### Query Parameters
+
+* `cascade` - _Optional_. The comma separated alternative routing systems to which the lookup is cascaded in addition to searching IPNI index records. If unspecified, only IPNI index records are searched. Supported values:
+    * `ipfs-dht`- The IPFS Kademlia DHT.
+
 ##### Response
 
 * `application/json` - JSON encoded [`FindResponse`][find-response-schema]. See [JSON Find Response](#json-find-response) example.
@@ -487,6 +492,12 @@ Given a multihash as path parameter, returns a list of its content providers.
 ##### Path Parameters
 
 * `multihash` - _Required_. The Base58 string representation of multihash value.
+
+##### Query Parameters
+
+* `cascade` - _Optional_. The comma separated alternative routing systems to which the lookup is cascaded in addition to searching IPNI index records. If unspecified, only IPNI index records are searched. Supported values:
+  * `ipfs-dht`- The IPFS Kademlia DHT.
+
 
 ##### Response
 
@@ -519,6 +530,12 @@ Base58 string representation. See [`FindRequest`][find-request-schema] schema.
   ]
 }
 ```
+
+##### Query Parameters
+
+* `cascade` - _Optional_. The comma separated alternative routing systems to which the lookup is cascaded in addition to searching IPNI index records. If unspecified, only IPNI index records are searched. Supported values:
+    * `ipfs-dht`- The IPFS Kademlia DHT.
+
 
 ##### Response
 
--- a/IPNI.md
+++ b/IPNI.md
@@ -459,6 +459,22 @@ to fetch the advertisement chain.
 
 An indexer node can be queried over HTTP for a multihash or a CID. This section provides a summary of the HTTP query APIs. A full OpenAPI specification of the APIs can be found [here](schemas/v1/openapi.yaml).
 
+
+#### Cascading Lookup
+
+The HTTP query API supports cascading queries for a given multihash or CID onto alternative routing systems in addition to searching IPNI records.
+A client may optionally specify a query parameter with key set to `cascade`, and value set to comma separated alternative routing systems, which are also searched for records.
+
+The specification imposes no constraints on the order by which the results are returned.
+Implementers are free to return results as they are found.
+
+The alternative routing systems currently supported is:
+ * `ipfs-dht`: equivalent to searching records on the IPFS network.
+
+A client may discover the list of alternative routing systems supported by a server via sending `OPTIONS` request.
+In response, the server may include `X-IPNI-Allow-Cascade` header key, with value as the comma separated list of alternative routing systems supported.
+The absence of this header key implies that the server does not offer cascading lookups.
+
 #### `GET /cid/{cid}`
 
 Given a CID as path parameter, returns a list of its content providers. The lookup ignores CID codec
@@ -471,8 +487,7 @@ and uses the multihash portion of the CID only.
 
 ##### Query Parameters
 
-* `cascade` - _Optional_. The comma separated alternative routing systems to which the lookup is cascaded in addition to searching IPNI index records. If unspecified, only IPNI index records are searched. Supported values:
-    * `ipfs-dht`- The IPFS Kademlia DHT.
+* `cascade` - _Optional_. The comma separated alternative routing systems to which the lookup is cascaded in addition to searching IPNI index records. If unspecified, only IPNI index records are searched. See [Cascading Lookup](#cascading-lookup)
 
 ##### Response
 
@@ -495,8 +510,7 @@ Given a multihash as path parameter, returns a list of its content providers.
 
 ##### Query Parameters
 
-* `cascade` - _Optional_. The comma separated alternative routing systems to which the lookup is cascaded in addition to searching IPNI index records. If unspecified, only IPNI index records are searched. Supported values:
-  * `ipfs-dht`- The IPFS Kademlia DHT.
+* `cascade` - _Optional_. The comma separated alternative routing systems to which the lookup is cascaded in addition to searching IPNI index records. If unspecified, only IPNI index records are searched. See [Cascading Lookup](#cascading-lookup)
 
 
 ##### Response
@@ -533,8 +547,7 @@ Base58 string representation. See [`FindRequest`][find-request-schema] schema.
 
 ##### Query Parameters
 
-* `cascade` - _Optional_. The comma separated alternative routing systems to which the lookup is cascaded in addition to searching IPNI index records. If unspecified, only IPNI index records are searched. Supported values:
-    * `ipfs-dht`- The IPFS Kademlia DHT.
+* `cascade` - _Optional_. The comma separated alternative routing systems to which the lookup is cascaded in addition to searching IPNI index records. If unspecified, only IPNI index records are searched. See [Cascading Lookup](#cascading-lookup)
 
 
 ##### Response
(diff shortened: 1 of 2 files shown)


masih:
@aschmahmann I would love your review on this spec change please

willscott:
also probably relevant for @lidel

ischasny (approved):

IPNI.md
```diff
@@ -463,6 +463,11 @@ and uses the multihash portion of the CID only.
 * `cid` - _Required_. The default string representation of the Cid. Currently, Base32 is used for
   CIDv1 as the encoding for the multibase string and Base58 is used for CIDv0.
 
+##### Query Parameters
+
+* `cascade` - _Optional_. The comma separated alternative routing systems to which the lookup is cascaded in addition to searching IPNI index records. If unspecified, only IPNI index records are searched. Supported values:
+    * `ipfs-dht`- The IPFS Kademlia DHT.
```
lidel, line 469: I assume you mean DHT created by self-[identifying](https://github.com/libp2p/specs/tree/master/identify#protocols) peers that speak `/ipfs/kad/1.0.0` libp2p protocol?

Mind adding clarification, which DHT you mean? (Kubo has at least two, we may have more in the future when if double hashing is exposed as separate DHT).

IPNI.md
```diff
@@ -463,6 +463,11 @@ and uses the multihash portion of the CID only.
 * `cid` - _Required_. The default string representation of the Cid. Currently, Base32 is used for
   CIDv1 as the encoding for the multibase string and Base58 is used for CIDv0.
 
+##### Query Parameters
+
+* `cascade` - _Optional_. The comma separated alternative routing systems to which the lookup is cascaded in addition to searching IPNI index records. If unspecified, only IPNI index records are searched. Supported values:
```
davidd8, line 468: To me, the term "cascade" implies a sequential order. Does that mean an IPNI indexer would wait for an initial response from the IPNI index, and if no results, go on to the next routing system, or ask all routing systems regardless of the results?

If the latter, it may be more clear to change the terminology to "routes" or "routing networks". Otherwise, let's clarify if order matters, or not.

IPNI.md
masih: > I assume you mean DHT created by self-[identifying](https://github.com/libp2p/specs/tree/master/identify#protocols) peers that speak /ipfs/kad/1.0.0 libp2p protocol?

Good question. The answer is: Sort of. Allow me to explain:

The idea here is to provide _labels with implicit meaning_ instead of concrete protocol IDs.
The mapping between the two _could be 1:1_; but not necessarily. It offers a user-friendly flex point in the protocol where we can use as a tuning knob, adding labels and deprecating exiting ones to fit what's contextually needed by the users.

In this case what is meant by the label `ipfs-dht` could be one or more protocol IDs. For example, if/when the double hashing DHT materialises the label would span to automatically include the new DHT protocol ID and "just work" by inferring which specific DHT it should do lookups in for a given multihash by checking the multihash code. 

What are your thoughts? Does that make sense?
Would it be clearer if it's changed to `ipfs` or `kubo`?

IPNI.md
masih: > Does that mean an IPNI indexer would wait for an initial response from the IPNI index, and if no results, go on to the next routing system, or ask all routing systems regardless of the results?

The specification here is specifically loose on the strategy. It is left to the implementers to decide on performing sequential lookup, only carry on if no result is found, find in parallel, etc. This offers flexibility depending on the implementations and whatever SLOs they strive for.

I understand the word "cascade" implies order. What it really means here is "also look up in".  "cascade" was the only single word I could think of that was close enough which was also used in discussions with folks. I also think regardless of order, "cascade" fits. Because the specification states that whether the parameter is specified or not, IPNI is always searched and any found results are included. 

In practice results will most likely be streamed back as they are found as long as the response timeout allows it (i.e. `ndjson` streaming that is planned to be implemented across HTTP delegated routing and cid.contact). Because, lookup in different routing systems has different latency characteristics. If order becomes an issue we can later introduce `order` query parameter that further narrows down the search strategy preference.

IPNI.md
```diff
@@ -463,6 +463,11 @@ and uses the multihash portion of the CID only.
 * `cid` - _Required_. The default string representation of the Cid. Currently, Base32 is used for
   CIDv1 as the encoding for the multibase string and Base58 is used for CIDv0.
 
+##### Query Parameters
```
aschmahmann, line 466: What's the rationale for implementing this in the IPNI-specific API rather than as an extension to https://github.com/ipfs/specs/pull/337 (which IIUC is the thing meant to be used by external consumers of multiple routing systems such as kubo, helia, ...)? I guess it could be in both, but given that the semantics of different routing systems could be different it seems reasonable to me that IPNI would have a protocol more specifically tied to what it does, and a more generic API would cover plugging in additional routing systems.

Putting this in the non-IPNI API would also mean the IPNI API doesn't need to care about streaming results and such if it doesn't want to.

IPNI.md
aschmahmann, line 468: Allowing for multiple parameters without the user being able to know which are available or were used to satisfy a request still leaves implementations a little clueless about what to do.

For example, if I don't know if cid.contact has an IPFS Public DHT plugin and I query it how will I know if I should also search the IPFS Public DHT?

It seems like in order for this to work that you'd want a "what do I speak" endpoint. You might also want to tag responses with the set of routers you checked.

IPNI.md
```diff
@@ -463,6 +463,11 @@ and uses the multihash portion of the CID only.
...
+
+* `cascade` - _Optional_. The comma separated alternative routing systems to which the lookup is cascaded in addition to searching IPNI index records. If unspecified, only IPNI index records are searched. Supported values:
+    * `ipfs-dht`- The IPFS Kademlia DHT.
+
 ##### Response
```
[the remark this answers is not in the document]
masih: https://github.com/ipni/specs/issues/8

IPNI.md
masih: Great point. 

There are two flavours of HTTP APIs: IPNI and the new HTTP delegated routing (ignoring the retired `/reframe` endpoints)

IPNI flavour predates the others. We need to maintain it for compatibility until, at the very least, HTTP delegated routing is stable.

Changing IPNI specs internally to what's needed has a faster turnaround and less resistance compared to the HTTP delegated routing. Therefore, the strategy here is to do both: try things in IPNI API first and see how they fit, then propose changes in HTTP delegated routing.

IPNI.md
aschmahmann: > Today, the ask is to optionally search the DHT without doubling traffic on the DHT when requests come from Kubo. I think the proposed changes here deliver on that.

Technically this is correct even if cascading was mandatory since Kubo doesn't use this API, but instead the `/routing/v1` API (the fact that the indexstar implementation of `/routing/v1` uses IPNI APIs to backend implementations internally is somewhat coincidental. However, I get your point that this does the minimal thing.

---

What I suspect you'd want to enable here is that developers (or users) could choose to forgo public DHT requests entirely and just send them all to cid.contact if they wanted, and only do the requests themselves if cid.contact was failing/unavailable for some reason. In order for them to do this in the current proposal, they'd need to hard-code "hey, cid.contact speaks dht" somewhere in their code/config.

This is the same kind of thing that is likely to cause user problems in https://github.com/ipfs/specs/pull/342.

IPNI.md
aschmahmann: > Changing IPNI specs internally to what's needed has a faster turnaround and less resistance compared to the HTTP delegated routing.

In this particular case I don't really understand why though. Basically you can choose to put the DHT lookup node in one of three configurations (note: I'm using domain names to roughly describe the layout, what the actual domain names are called, which group runs them, ... 🤷)
1. User -> cid.contact -> (IPNI-DB lookup + DHT lookup)
2. User -> routing.delegate.ipfs.io -> (DHT lookup + cid.contact)
3. User -> routing.delegate.ipfs.io -> (dht.delegate.ipfs.io + cid.contact)

Any of these options will get the job done. This approach is option 1. The reasons I personally prefer options 2 or 3 are:
1. It makes it harder to figure out what an IPNI endpoint does and what to do with it (i.e. my comments in this thread https://github.com/ipni/specs/pull/9#discussion_r1088968679)
2. It seems to cause tension when there is need for some capability in a routing system, that IPNI may/may not be interested in proxying (e.g. this IPNS comment https://github.com/ipfs-shipyard/someguy/issues/13#issuecomment-1406478765 and related comments in that issue)
   - There is a need for one (or a set) of delegated routing endpoints that low-resource clients (e.g. browsers) can leverage. Is that cid.contact, or something else, idk. But it tends to put pressure in this area.

Perhaps options 2/3 increase operational costs associated with caching in cloud infra compared to option 1, or there's some other good reason to go with option 1. If the IPNI team feels strongly that option 1 is the best way forward then 👍, I mostly just ask that my two concerns above be taken into consideration as we move things forward.

masih: No strong feelings here :)

Those options look very similar to me. I think no matter where we put this i'd like to point out:
* we will need some specification for it (i.e. the likes of this PR).
* this spec does not change the default cid.contact behaviour; the default response will include IPNI records only.
* considering cid.contact heavily utilises caching, and is already set up/running it seems like the path of least resistance to place this functionality in option 1 and take advantage of the caching.
* as you pointed out this will have cost saving benefits. 
* assuming the other endpoints listed above do not exist today, it'd also be the fastest path to make the service available.
* regarding [this thread](https://github.com/ipni/specs/pull/9#discussion_r1088968679), would [fa14d37](https://github.com/ipni/specs/pull/9/commits/fa14d372a34e5151dbb009f3b43f4733e7eb3e8d) address your concern?

IPNI.md
masih: Added extra paragraphs to spell out the absence of order constraints.

IPNI.md
lidel: No, there is nothing specific to  `kubo` in `/ipfs/kad/1.0.0`.

> The idea here is to provide labels with implicit meaning instead of concrete protocol IDs. 

Ack, then `ipfs-dht` is fine :+1:  Thanks for explanation!

IPNI.md
aschmahmann: Yep, that addresses the concern thanks 🙏. As for where we should shuffle the query endpoints 🤷, to be tackled another day.

masih: Thank you @aschmahmann

PR #14, opened 2023-04-04
Document cascading lookup over legacy bitswap broadcast

masih:
Update the specification with extension to legacy cascade over BitSwap along with options to discover peering addrs.

### DIFF
--- a/IPNI.md
+++ b/IPNI.md
@@ -470,10 +470,18 @@ Implementers are free to return results as they are found.
 
 The alternative routing systems currently supported is:
  * `ipfs-dht`: equivalent to searching records on the IPFS network.
+ * `legacy`: searches records via BitSwap broadcast over a set of peered nodes; see [Legacy Lookup Cascade](#legacy-lookup-cascade).
 
 A client may discover the list of alternative routing systems supported by a server via sending `OPTIONS` request.
 In response, the server may include `X-IPNI-Allow-Cascade` header key, with value as the comma separated list of alternative routing systems supported.
 The absence of this header key implies that the server does not offer cascading lookups.
+
+##### Legacy Lookup Cascade
+
+The legacy lookup cascade propagates search for providers via BitSwap broadcast protocol across a list of nodes that have explicitly peered with the legacy cascade endpoint. 
+A client may discover the the legacy cascade endpoints provided by a server by sending an `OPTIONS` request.
+In response, the server may include `X-IPNI-Legacy-Cascade-Peering` header key, with value as a multiaddr to which to peer connections.
+There may be multiple peering multiaddrs offered by a server, in which case the header key may be repeated once for each peering address.
 
 #### `GET /cid/{cid}`
 
--- a/schemas/v1/openapi.yaml
+++ b/schemas/v1/openapi.yaml
@@ -287,6 +287,13 @@ components:
             The absence of this header implies that the server does not offer cascading lookups.
           schema:
             type: string
+        X-IPNI-Legacy-Cascade-Peering:
+          description: |
+            The legacy cascade multiaddr to which to peer in order to make content provided only over Bitswap discoverable by an IPNI instance.
+            This header may be repeated multiple times when the server offers more than one peering address.
+            The absence of this header implies that the server does not offer legacy cascade peering.
+          schema:
+            type: string
 externalDocs:
   description: IPNI Specification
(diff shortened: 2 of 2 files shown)


IPNI.md
```diff
@@ -470,11 +470,18 @@ Implementers are free to return results as they are found.
...
 
 A client may discover the list of alternative routing systems supported by a server via sending `OPTIONS` request.
 In response, the server may include `X-IPNI-Allow-Cascade` header key, with value as the comma separated list of alternative routing systems supported.
 The absence of this header key implies that the server does not offer cascading lookups.
 
+##### Legacy Lookup Cascade
+
+The legacy lookup cascade propagates search for providers via BitSwap broadcast protocol across a list of nodes that have explicitly peered with the legacy cascade endpoint. 
+A client may discover the the legacy cascade endpoints provided by a server via sending `OPTIONS` request.
+In response, the server may include `X-IPNI-Legacy-Cascade-Peering` header key, with value as the comma separated list of multiaddrs to which to peer connections.
```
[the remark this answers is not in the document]
masih: One header per multiaddr is probably more readable 👍

masih:
Cc @aschmahmann @lidel @guseggert

IPNI.md
```diff
@@ -470,11 +470,19 @@ Implementers are free to return results as they are found.
...
+ * `legacy`: searches records via BitSwap broadcast over a set of peered nodes; see [Legacy Lookup Cascade](#legacy-lookup-cascade).
 
 A client may discover the list of alternative routing systems supported by a server via sending `OPTIONS` request.
 In response, the server may include `X-IPNI-Allow-Cascade` header key, with value as the comma separated list of alternative routing systems supported.
 The absence of this header key implies that the server does not offer cascading lookups.
 
+##### Legacy Lookup Cascade
+
+The legacy lookup cascade propagates search for providers via BitSwap broadcast protocol across a list of nodes that have explicitly peered with the legacy cascade endpoint. 
+A client may discover the the legacy cascade endpoints provided by a server via sending `OPTIONS` request.
```
gammazero, line 482:
```suggestion
A client may discover the the legacy cascade endpoints provided by a server by sending an `OPTIONS` request.
```

gammazero (approved):

IPNI.md
```diff
@@ -470,11 +470,19 @@ Implementers are free to return results as they are found.
 
 The alternative routing systems currently supported is:
  * `ipfs-dht`: equivalent to searching records on the IPFS network.
+ * `legacy`: searches records via BitSwap broadcast over a set of peered nodes; see [Legacy Lookup Cascade](#legacy-lookup-cascade).
```
lidel, line 473: :thought_balloon: a meta-comment on assuming things will never change ;-) 

`legacy` is a bit unfortunate name, because in 2 years something else will be considered "legacy".

Something like `bitswap-peering` would be way better and more future-proof.

schemas/v1/openapi.yaml
```diff
@@ -287,6 +287,13 @@ components:
             The absence of this header implies that the server does not offer cascading lookups.
           schema:
             type: string
+        X-IPNI-Legacy-Cascade-Peering:
```
lidel, line 290: :thought_balloon:   "Legacy Cascade"  sounds like a noun of some protocol, while it is just an implementation detail of IPNI doing fallback to Bitswap Peering.

IPNI.md
masih: The term legacy is there to convey that this functionality will be deprecated.

achingbrain: Deprecation could be mentioned in a comment?

If the routing method is bitswap peering, it makes sense to call it that since then you need no other context to know what it is.

willscott:
* I think the "bitswap-peering" is going to be a bit awkward as a cascade because it's unclear which peers are being peered with by a given provider.
* I wonder if it would make sense to call it something like "infura-bitswap-peering" or "2023-non-dht-bitswap-peering" or something to express the caveat
* I agree with @masih that it seems reasonable at this point to hope it doesn't last for particularly long, and is not a mechanism i'd expect to encourage other partners to run

IPNI.md
masih: Thank you for pointing this out; deprecation notice added here: https://github.com/ipni/specs/pull/15

In terms of naming the one thing i wanted all of this to convey is that cascading over bitswap over peering is not here to stay and is "legacy" already.

PR #18, opened 2023-05-23
Document IPNI HTTP provider specification

### LINKED ISSUE #16: Formalize ipni advertisement publishing http spec
currently this is the behavior described in https://github.com/ipni/specs/blob/main/IPNI.md#http-1

we probably want to express a specific implicit path we might want as a root for the advertisement chain over http.

This would be a more specific doc targeted at defining what a provider would need to do to be considered an 'http index provider'

masih:
[Rendered](https://github.com/ipni/specs/blob/masih/http_prov_spec/IPNI_HTTP_PROVIDER.md)

**Note: I have opted for least disruptive spec relative to the current implementation. If you folks think this doesn't matter as much there are things that I'd like to change.**

Fixes #16


willscott (approved):
this looks like a good extraction of what we expect from the provider

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,284 @@
...
+```
+
+The following snippet shows an example of advertisement entries encoded in `DAG-JSON`: 
+
+```json
+{
+  "Entries": [
+    {
+      "/": {
+        "bytes": "EiDYa/LrNYDdHjSlKGXpHjc1IyHds9Xdi/p8d25Q5UE/bQ=="
```
rvagg, line 234:
```suggestion
        "bytes": "EiDYa/LrNYDdHjSlKGXpHjc1IyHds9Xdi/p8d25Q5UE/bQ"
```

(these padded examples need to disappear into history)

gammazero (approved):
This is excellent

rvagg:
It might be nice to leave open (and document here) the ability of announcement `Addrs` to include optional [`httpath`](https://github.com/ipni/go-libipni/blob/1bd7b3f90d2a849c3449742947e86d78e66508ab/maurl/convert.go#L35-L41) specifier, and be clear how this interacts with the new path prefixes defined here. If I provide an `httpath` in a multiaddr, does it override `/ipni/v1/ad/`, or does it prepend to this?
My own limited use of this has involved wanting to change the default provider server that responds to `/head` and `/{cid}` to one that responds on `/_ipni/head` and `/_ipni/{cid}` because I want to [reuse an existing server that has other things going on at the root](https://github.com/ipni/go-libipni/pull/34). So I like that this is now namespaced; I also like optionality, but it's not critical now that this is so specific. i.e. custom `httpath` could be ruled out in this spec and I wouldn't be too upset, but I'd like the clarity.

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,284 @@
...
+  },
+  "sig": {
+    "/": {
+      "bytes": "LrZiicKdqDqkG2UFR..."
+    }
+  }
+}
+```
+
+### GET `/ipni/v1/ad/{CID}`
```
ischasny, line 184: Would it be helpful to have bulk fetches in the specification? Something like `GET /ipni/v1/ad/{CID}?depth=20`, so that chains can be traversed faster. Support for depth can be optional too.

masih: Out of scope for this spec IMO. 

How it occurs for me is that future iterations can introduce this sort of functionality, along with support for `[application/vnd.ipld.car](https://www.iana.org/assignments/media-types/application/vnd.ipld.car)` where we can ask a provider to preassemble CAR files containing ads.

This functionality is sufficiently complex that deserves its own spec/iteration IMHO. 

Cc @gammazero

ischasny: Sounds good, this is more of a optimisation rather than a must have. Other than that - looks good to me.

ischasny (approved):

ischasny (approved):

IPNI_HTTP_PROVIDER.md
masih: Added this as an item in [future considerations](https://github.com/ipni/specs/blob/masih/http_prov_spec/IPNI_HTTP_PROVIDER.md#future-considerations)

rvagg:
great! I've followed up in https://github.com/ipni/go-libipni/issues/42

masih:
Cc @lidel @aschmahmann

BigLep:
This came up in https://pl-strflt.notion.site/2023-05-30-Content-Routing-WG-12-b2ed74834fe44e359bbcdd02740e2084

There is going to be implementation need for this in the next quarter or two.  As a result, want to get ahead before code gets written and decisions ossify.

Some next steps:
1. Update the IPIP to the newest template standard with lessons learned the last 6 months
2. Review from spec maintainers

Implementation notes:
"Lassie currently does not depend on boxo. It may be hard for us to prototype ambient-discovery in boxo until the boxo/rest-of-world dependency conflicts are resolved"

(I'll come through and update this more later today.)

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,310 @@
...
+
+1. **Explicit Definition** -- A formal definition of the HTTP API for IPNI providers fills a crucial gap in the IPNI
+   specification. It provides a comprehensive and explicit outline of the required endpoints, request/response formats,
+   and behavior expected from providers. Such definition can then be referenced as part of the capability required by a
+   provider that aims to offer retrieval over HTTP.
+
+2. **Grouping of Capabilities** -- An implicit grouping of API endpoints is essential because HTTP servers running IPNI
+   capabilities often provide multiple functionalities. By organizing the API endpoints under a unified specification,
+   it becomes easier to understand and navigate the various capabilities offered by an IPNI provider while avoiding
+   conflict with other APIs such as the IPFS Gateway or Boost.
```
BigLep, line 70:
```suggestion
   conflict with other APIs such as the [IPFS HTTP Gateway](https://specs.ipfs.tech/http-gateways/) or Boost.
```

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,310 @@
...
+   of the IPNI ecosystem, while reducing engineering footprint.
+
+Overall, the formal definition of the HTTP API for IPNI providers enhances clarity, promotes consistency, and opens up
+opportunities for innovation and collaboration within the IPNI network.
+
+## Background
+
+### Common Data Types
+
+The IPNI HTTP Provider API specification makes use of the following common data types:
```
BigLep, line 84: Your call, but if you're expecting others outside PL to read this, could be good to link to sources for more information.

Maybe use reference style links?

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,310 @@
...
+
+#### Implicit Namespace Assumption
+
+Indexer nodes implicitly assume the `/ipni/` URL prefix. When receiving an announcement from providers, an indexer will attempt to fetch advertisements by joining the announced URL with the paths documented in this specification. Implementers have the freedom to include additional paths to the announced URL if desired.
+
+For example, if the announced URL is `https://ipni-provider.example`, it will be accessed at `https://ipni-provider.example/ipni/v1/ad/head` to fetch the head advertisement. Similarly, if the announced URL is `https://ipni-provider.example/my/prefix`, it will be accessed at `https://ipni-provider.example/my/prefix/ipni/v1/ad/head`.
+
+#### Custom Paths in Multiaddr
+
+IPNI implementers have the option to announce HTTP publisher URLs in `Multiaddr` format. However, this format does not officially support paths. To accommodate custom paths in the announced URLs, implementers can utilize the `httpath` parameter. The handling of `httpath` is similar to that of vanilla HTTP URLs with an additional specified path.
```
BigLep, line 143: 1. (newbie question) Is `httpath` something we're coming up with now, or is there existing discussion on this?
2. Is this intentional to be `httpath` rather than `httppath` ?

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,310 @@
...
+### Security
+
+The IPNI HTTP Provider API specification includes several security implications and considerations to ensure secure
+communication and data integrity. Here are the key security aspects:
+
+- **TLS (Transport Layer Security)**: It is recommended to use TLS whenever possible to establish secure connections
+  between clients and the IPNI HTTP provider. TLS encrypts the data during transmission, protecting it from unauthorized
+  access and tampering.
+
+- **Pagination and Maximum Content Length**: Implementers should consider implementing pagination mechanisms and setting
```
BigLep, line 272: Don't you want to standardize how paging is communicated between clients and servers to avoid divergence between implementations?

IPNI_HTTP_PROVIDER.md
BigLep: Oh I learned about https://github.com/ipni/go-libipni/issues/42 at the end of the doc.  This would be another reason to use reference links.  I would have learned about it right away.

BigLep (commented):
Don't block on anything from me.  Just left comments while reading through after it came up during the Content Routing WG on 2023-05-30.  Thanks.

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,310 @@
...
+
+For example, if the announced URL is `https://ipni-provider.example`, it will be accessed at `https://ipni-provider.example/ipni/v1/ad/head` to fetch the head advertisement. Similarly, if the announced URL is `https://ipni-provider.example/my/prefix`, it will be accessed at `https://ipni-provider.example/my/prefix/ipni/v1/ad/head`.
+
+#### Custom Paths in Multiaddr
+
+IPNI implementers have the option to announce HTTP publisher URLs in `Multiaddr` format. However, this format does not officially support paths. To accommodate custom paths in the announced URLs, implementers can utilize the `httpath` parameter. The handling of `httpath` is similar to that of vanilla HTTP URLs with an additional specified path.
+
+For example, if the announced multiaddr is `/dns/ipni-provider.example/https`, it will be accessed at `https://ipni-provider.example/ipni/v1/ad/head` to fetch the head advertisement. On the other hand, if the announced multiaddr is `/dns/ipni-provider.example/https/httpath/%2Fmy%2Fprefix`, it will be accessed at `https://ipni-provider.example/my/prefix/ipni/v1/ad/head`.
```
MarcoPolo, line 145: I've left a comment here: https://github.com/ipni/go-libipni/issues/42#issuecomment-1569330073. More context there, but I don't think we should be putting application logic in the multiaddr.

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,310 @@
...
+The IPNI Provider HTTP API adopts a versioning strategy that utilizes the path prefix to indicate major releases. As a
+best practice, future iterations that introduce breaking changes should increment the version number in the API path.
+This approach ensures compatibility and allows clients to adapt to new versions while maintaining backward compatibility
+with existing implementations. By explicitly incorporating versioning in the API path, it becomes easier to manage and
+track changes over time, facilitating smooth transitions and providing a clear indication of API evolution.
+
+
+### Namespacing
+
+This specification introduces the `/ipni/` URL prefix to namespace the functionality associated with IPNI HTTP providers. The presence of this namespace allows IPNI specifications to evolve without conflicting with other URL patterns that an HTTP server may support. Therefore, this specification reserves the right for any sub-tree that appears after this prefix.
```
lidel, line 133: :thought_balloon:  Is the long term plan  to expose `/cid` and `/multihash`  to `/ipni/v1` namespace too?

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,310 @@
...
+
+IPNI implementers have the option to announce HTTP publisher URLs in `Multiaddr` format. However, this format does not officially support paths. To accommodate custom paths in the announced URLs, implementers can utilize the `httpath` parameter. The handling of `httpath` is similar to that of vanilla HTTP URLs with an additional specified path.
+
+For example, if the announced multiaddr is `/dns/ipni-provider.example/https`, it will be accessed at `https://ipni-provider.example/ipni/v1/ad/head` to fetch the head advertisement. On the other hand, if the announced multiaddr is `/dns/ipni-provider.example/https/httpath/%2Fmy%2Fprefix`, it will be accessed at `https://ipni-provider.example/my/prefix/ipni/v1/ad/head`.
+
+By supporting custom paths through `httpath`, IPNI implementers can enhance the flexibility of their HTTP publisher URLs while maintaining compatibility with the IPNI HTTP Provider API.
+
+## Specification
+
+### GET `/ipni/v1/ad/head`
```
lidel, line 151: I assume this either has a limited TTL where a stale response can be cached for some time,  or should not be cached at all, so the client gets up-to-date response.

Either way, you should require implementer to always return explicit `Cache-Control`:

- To disable HTTP caching and always fetch the latest response:
  ```
  Cache-Control: no-cache, no-store, must-revalidate
  ```
- To allow accepting a cached response and revalidating it in the background, you can use the following Cache-Control header:
   ```
   Cache-Control: max-age=60, stale-while-revalidate=120
   ```
   if the cached response is stale (older than 60 seconds), the client can still serve it while simultaneously sending a request to the server to get updated one -- this removes spikes in latency when Cloudfront  experiences cache miss due to cache age

lidel (commented):
@masih added my usual comments about including explicit headers for HTTP Caching + some questions/concerns around `/httpath`

IPNI_HTTP_PROVIDER.md
masih: > for endpoint of HTTP-only(?) service,

The service may be using vanila TCP or libp2p for transport.

IPNI_HTTP_PROVIDER.md
masih: This is a provider spec only. the `cid` and `multihash` endpoints are for looking up records which reside on the server side and are outside of the scope of this doc.

IPNI_HTTP_PROVIDER.md
masih: "application" is a loose term. From my perspective this is not an application logic in the same way that in a web 2 world URL to a REST API is not an application logic. It's a standard configuration for HTTP API interaction.

IPNI_HTTP_PROVIDER.md
masih: Pagination is achieved by IPLD links across entry chunks. I'll add a paragrah to spell this out. Thanks Steve.

IPNI_HTTP_PROVIDER.md
masih: Removing support for `httpath` in multiaddr as per Slack discussion.

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,310 @@
...
+  },
+  "sig": {
+    "/": {
+      "bytes": "LrZiicKdqDqkG2UFR..."
+    }
+  }
+}
+```
+
+### GET `/ipni/v1/ad/{CID}`
```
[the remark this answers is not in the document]
masih: Added notes on caching across the doc. Thanks @lidel

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,328 @@
...
+
+```
+Cache-Control: public, max-age=29030400, immutable
+```
+
+By setting this response header, it instructs client-side caches and intermediate proxies to store the response for a long duration (`max-age=29030400`), as well as treat it as immutable. This helps improve performance and ensures that the same content is served consistently for the specified `CID`.
+
+### Response
+
+The response from the `/ipni/v1/ad/{CID}` endpoint returns the data associated to the advertisement CID and its embedded links, including `Entries`. See [IPNI Specification/Advertisements](IPNI.md#advertisements) for the advertisement and entries schema.
```
gammazero, line 233: It is not clear whether:
1. The entries are included in the response to a request for an advertisement CID
2. One separate request for all the entries is required (unless max content length exceeded)
3. A separate request is required for each block of entries.

Option 2 is the best choice IMO.  Not 1, because a removal advertisement later in the chain may mean that the indexer does not need to download the entries. Also, a utility inspecting the chain may not want the entries. Not 3, because the provider should be able to stream these to the indexer (nd-json encoded).

It should also be stated explicitly that the same endpoint is used to request advertisements as well as entries blocks.

Suggest:
> Entries must be retrieved using a separate request for the CID of the next entries block to fetch. The response contains one or more blocks of entries, up to the maximum content length. The first block in the response must match the CID requested (see entries verification). If more entries remain to be fetched, this is indicated by the last included entries block having a non-empty "Next" field. A subsequent request for the CID specified in the "Next" field retrieves the next set of entries blocks. This continues until the last block has an empty "Next" field.

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,328 @@
...
+
+- **Signature Validation**: Implementers should validate the signatures associated with the head advertisement CID to
+  verify the authenticity and integrity of the data. Signature validation ensures that the data is generated by the
+  legitimate provider and has not been tampered with during transmission.
+
+- **Content Verification against CID**: Implementers should verify that the content associated with an advertisement or
+  entry CID matches the expected CID value. This verification ensures the consistency and accuracy of the data, guarding
+  against any manipulation or corruption.
```
gammazero, line 301: Need to specify exactly how to verify an entry:

The CID of each entry chunk can be verified by creating a `schema.EntryChunk` with its `Entries` set to the array of multihashes from the response chunk to verify and the `Next` field set to the CID of the response chunk's `"Next"` field. Then use the IPLD `LinkSystem.ComputeLink` to compute the CID for the `schema.EntryChunk`.

Since the advertisement signature covers the CID of the first entry chunk, verifying the CIDs of the entry chunks also verifies the signature of all the chunks. This works because changing a CID in a chunk alters its CID, and altering the CID, and thus the link in the previous chunk, will change the CID of the previous chunk, all the way up to the first chunk that is covered by the advertisement signature.

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,328 @@
...
+
+```json
+{
+  "Entries": [
+    {
+      "/": {
+        "bytes": "EiDYa/LrNYDdHjSlKGXpHjc1IyHds9Xdi/p8d25Q5UE/bQ"
+      }
+    }
+  ]
```
gammazero, line 280: Add "Next" field.
```suggestion
  ],
  "Next": {
    "/": "baguqeeraog3uqu7au2mctrfslzvua5lzl3wyoquqhbyoig6hnputypej5guq"
  }
```

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,328 @@
...
+- **Writer Privacy**: Ensuring the privacy and confidentiality of the writers' data is an important aspect to address in
+  future iterations. Mechanisms such as encryption and access control can be explored to safeguard sensitive information
+  and protect the privacy of writers.
+
+- **Swappable Transport**: The IPNI HTTP Provider API can explore integration with
+  the [`go-libp2p-http`](https://github.com/libp2p/go-libp2p-http)  library to leverage its capabilities and enhance the
+  transport layer. This integration can enable seamless communication between IPNI providers and consumers using the
+  libp2p networking stack, offering additional features and benefits.
+
+- **Bulk Fetch**: The IPNI HTTP Provider API can introduce support for [`application/vnd.ipld.car`](https://www.iana.org/assignments/media-types/application/vnd.ipld.car) to allow bulk download of advertisements. Such extensions may include query parameters to further enable indexer nodes to refine and optimize the fetching process. See the [IPFS HTTP Gateway specification](https://specs.ipfs.tech/http-gateways/) for related work.
```
gammazero, line 316: Bulk fetch may be harmful when there are context ID removals. If a removal advertisement removes a context ID of an earlier advertisement, then there is no need to download the entries data for that advertisement. However it may still be necessary to download the advertisement to get the signature and link to the previous advertisement.

There should be some mechanism in a bulk fetch that allows only the optional retrieval of the advertisement portion of the data.

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,328 @@
...
+
+The IPNI HTTP Provider API specification includes several security implications and considerations to ensure secure
+communication and data integrity. Here are the key security aspects:
+
+- **TLS (Transport Layer Security)**: It is recommended to use TLS whenever possible to establish secure connections
+  between clients and the IPNI HTTP provider. TLS encrypts the data during transmission, protecting it from unauthorized
+  access and tampering.
+
+- **Pagination and Maximum Content Length**: Implementers should take into account setting a maximum content length for responses to mitigate potential abuse or resource exhaustion attacks. This helps ensure the API's performance and resilience against malicious actors. We suggest a maximum content length of 4MB. It is important to note that content pagination is achieved using IPLD links. For example, advertisement entry chunks are linked together using the `Next` field, forming a linked list of multihashes. Therefore, implementors must ensure that the generated entry chunks remain below the maximum content length accepted by the indexers.
+
```
gammazero, line 294: Each chunk can be whatever size the publisher of that data decides, as long as that size and any encoding overhead is below the maximum content length. Once a chunk is created, it cannot be resized since the CID of that chunk, as represented in the advertisement or previous chunks, must be verifiable.

IPNI_HTTP_PROVIDER.md
gammazero: I think that blocks should be capable of being streamed, using nd-json encoding. This would remove the need to paginate or to create hold a very large response body in memory.

IPNI_HTTP_PROVIDER.md
gammazero: If we want to have flexible chunk sizes while still being able to validate the data with the provider's signature, then the advertisement signature cannot include the first chunk CID and a separate signature is needed instead.

IPNI_HTTP_PROVIDER.md
masih: * Encoding of body is a separate issue and HTTP offers enough extensibility with Conent-Type and Accept headers to address it. 
* Entries are retrieved by their CID just like an advertisement via `/ipni/v1/ad/{CID}`. The word `ad` in the URI does not imply that it only returns IPLD nodes of type `Advertisement`. Instead it means the endpoint is dedicated to serving ad related content. This is stated [here](https://github.com/ipni/specs/pull/18/files#diff-499b31ce9fbe231ed9ab3dfc01be2bb456792a9edeb6f1df749e9b6c126db871R220).
* I don't think we want to have a separate URI for each IPNI model because to get a CID that one does not know the type of one would have to try all endpoints. It would also defeat the purpose of CIDs a bit?
* This would also solve the pagination problem in the same way that it is currently solved by `Next` link in entries traversal.

Does that make sense? 
In the meantime, I will refine the test to state that the endpoint returns all ad related content if [this](https://github.com/ipni/specs/pull/18/files#diff-499b31ce9fbe231ed9ab3dfc01be2bb456792a9edeb6f1df749e9b6c126db871R220) does not suffice.

IPNI_HTTP_PROVIDER.md
masih: By verification here we simply mean hash digest verification, in that the hash of content corresponds to the CID asked. Anything deeper goes into "ingest" territory IMHO which is out of the scope of this spec, since the spec is focused on the provider side.

IPNI_HTTP_PROVIDER.md
masih: This seems out of the scope of the spec.

IPNI_HTTP_PROVIDER.md
masih: Is there a specific requirement that we should satisfy which breaks the existing protocol? If not my sense is to push this out of the scope and leave this to be as is: i.e. recommendation on "consider max body sizes when chunking".

gammazero: The question is, do we want to allow the chunk size, and thus the chunk CID, to change without affecting the advertisement signature.

IPNI_HTTP_PROVIDER.md
gammazero: > Encoding of body is a separate issue

OK, the specifics of what is put into the body can go into a separate (OpenAPI) spec. 

> Entries are retrieved by their CID just like an advertisement via /ipni/v1/ad/{CID}

Thanks. Just wanted to clarify that.

> I don't think we want to have a separate URI for each IPNI model because to get a CID that one does not know the type of one would have to try all endpoints. It would also defeat the purpose of CIDs a bit?

The point of having a separate URI is to explicitly identify what type of CID is being asked for. If using the ad/ endpoint then the provider only needs to look in the place where ads are kept, and if using an entries/ endpoint then only look where the entries data is kept.  I cannot think of a time when the requestor will not know whether they are requesting an ad or an entry block.  

Anyway, this is OK to use the same endpoint for both. It just means that the provider will not know what kind of object the CID is retrieving, which is find. The provider just has to retrieve it.

 > This would also solve the pagination problem in the same way that it is currently solved by Next link in entries traversal.
 
 The question is about whether there is, or should be, any expectation about 1 entry block per request or about n entry blocks per request.

IPNI_HTTP_PROVIDER.md
gammazero: This should really be a question about **Signature Validation**.  We should be explicit about whether the advertisement signature covers the CID of the first entry block. I think the answer is yes, but it means that the entries cannot change size without changing the signature on the ads, which may be worth pointing out.

IPNI_HTTP_PROVIDER.md
gammazero: What that mechanism is, is out of scope. I fo think it is fundamental to advertisement retrieval, bulk or not, that fetching the entries must be optional, so thought it might be worth stating that. Happy to omit it if you think better to leave it out.

masih: I don't follow. Can you help me understand what you are asking to be changed here specifically?

IPNI_HTTP_PROVIDER.md
masih: Singature validation is covered [here](https://github.com/ipni/specs/pull/18/files#diff-499b31ce9fbe231ed9ab3dfc01be2bb456792a9edeb6f1df749e9b6c126db871R299). 

> Should we be explicit about whether the advertisement signature covers the CID of the first entry block?

It currently does and no where in the document we imply otherwise. What's the issue?

> Do we want to allow the chunk size, and thus the chunk CID, to change without affecting the advertisement signature?

Maybe in the future? Right now this seems out of the scope for this spec.

IPNI_HTTP_PROVIDER.md
```diff
@@ -0,0 +1,332 @@
...
+Custom paths in addresses specified as Multiaddr is not currently supported. See related issues below:
+* https://github.com/ipni/go-libipni/issues/42
+* https://github.com/libp2p/specs/pull/550
+
+## Specification
+
+### GET `/ipni/v1/ad/head`
+
+This endpoint retrieves the most recent advertisement CID published by the provider, along with additional information such as the provider's public key and the topic under which the advertisement is published.
```
gammazero, line 150: > topic under which the advertisement is published

Should this say "topic under which the advertisement is announced"?  If this is saying "published" then that seems to imply that the topic would be used to construct the URL path to retrieve the advertisement, but that is not stated anywhere.

gammazero (approved):

PR #27, opened 2023-12-19
Specify the IPNI federation protocol

masih:
Specify the initial IPNI federation protocol which aims to achieve eventually consistent index records across a collaborating set of nodes.

The federation protocol consists of four fundamental steps: Initialization, Periodic snapshot taking, Exchange of snapshots and Reconciliation. The protocol takes advantage of the immutability of advertisements exposed by each provider to resolve conflicts across indexers.

The specification lists a set of APIs exposed by a participating indexer in order to enable the implementation of the federation protocol.


IPNI_FEDERATION.md
```diff
@@ -0,0 +1,417 @@
...
+
+**Resolution Strategy:**  
+This represents a divergence in the advertisement chain, and careful reconciliation is required:
+
+1. Instance A and B determine which advertisement is the later one in the chain of advertisements provided by P1.
+2. They converge on the advertisement has been published the latest, i.e. has the shorter distance to P1's current 
+   advertisement head.
+
+While traversing advertisement chains, instances should prefer pulling ad chain from eachother's mirror instead of going
+directly to the provider where possible.
```
willscott, line 240: should we include the height of each head in snapshots, so that the understanding of which one is 'latest' is easier to resolve?

masih: We could but would that number be hearsay? I think height needs to be added to ad by provider, and signed instead but that's a much bigger change

Aside from trust, we also want to encourage indexers to fetch ad chains from each other's mirrors to replicate the chain. 

Thoughts?

willscott: one of the two parties should know the other parties head already since it will be older. A 'latest' provides a cheap hint for who is supposed to initiate the sync, but an un-trusting indexer could ignore it and still follow the protocol correctly.

willscott: it seems like an easy enough thing

masih: Also captured https://github.com/ipni/specs/issues/28

IPNI_FEDERATION.md
```diff
@@ -0,0 +1,406 @@
...
+* **Reconciliation**: Reconciliation is the process of ensuring that two or more IPNI instances have consistent index
+  records. It involves comparing the content of instances, identifying discrepancies, and making updates so that all
+  participating instances have the same content.
+
+* **Vector Clock (VC)**: Vector Clock is a mechanism used to determine the order of events in a distributed system. In
+  IPNI federation protocol, a vector clock helps track the sequence of index updates across multiple instances, thereby
+  assisting in maintaining the eventual consistency of the index.
+
+* **Snapshot**: A snapshot refers to a captured state of the index records at a particular point in time consisting of
+  the providers list known by the indexer, the latest processed advertisement CID for each provider and an epoch number.
```
gammazero, line 103: I think it would be useful to also have a link to the previous epoch. This would allow walking a chain by epoch, instead of having to iterate through each advertisement, if something deep in the chain needed to be accessed.

IPNI_FEDERATION.md
```diff
@@ -0,0 +1,406 @@
...
+
+This document provides an overview of the IPNI federation protocol: a protocol by which a set of IPNI instances
+collaborate in order to provide a unified eventually consistent view of the index records.
+
+## Introduction
+
+The Inter-Planetary Network Indexer (IPNI) offers a routing system that enables mass advertisement of content addressed
+data and lookup performance in order of milliseconds.
+This is achieved by a design where a single IPNI instance strives to maintain full network state knowledge.
+Many IPNI instances can be instantiated across the globe to offer fast local ingestion and lookup of advertised content.
```
gammazero, line 28: Another goal of federation may be to allocate portions of the index content to different indexers. I think this should be talked about - even if only to say why we do not do it. People will want to index some stuff for some time, but not everything forever, and will want to be part of a federation for some of the other features.

Having local fast lookup is a good goal, but does that apply to everything? It may not be necessary to be fast and local for all storage providers, but just the highly accessed ones. This means we may want to consider a strategy that allows some subset of service providers, and even epochs within those providers' chains, to be assigned to various indexers. Infrequently accessed storage provider content can be indexed by fewer indexers.

Such a strategy may also be useful so that indexers can choose which providers and epochs are the best to index and not have to index everything, even from specific provider(s).

IPNI_FEDERATION.md
```diff
@@ -0,0 +1,406 @@
...
+data from a snapshot, it updates its index accordingly. Through continuous reconciliation processes, the protocol
+ensures that all IPNI instances trend towards eventual consistency.
+
+### Conflict Resolution
+
+In the vast landscape of the Inter-Planetary Network Indexer (IPNI), advertisement chains serve as the core mechanism
+for content propagation. As multiple IPNI instances exist across the globe, the potential for discrepancies in the state
+of these chains is undeniable. It becomes essential, therefore, to devise strategies that can reconcile these
+discrepancies effectively.
+
```
gammazero, line 190: Is there a use case to check that providers are indexing what they claim to index?  An indexer may try to only index popular, dropping data it does not get any hits on after some amount of time. Another indexer (or index checker) could prove that an indexer is not indexing some content that it should be, and that could result in some penalty or reputation damage.

IPNI_FEDERATION.md
```diff
@@ -0,0 +1,406 @@
...
+1. Initialization - the process by which a node joins a federated network.
+2. Periodic snapshots - the process by which a node periodically captures an immutable snapshot of its providers along
+   with associated metadata.
+3. Snapshot exchange - the process by which each node propagates their latest snapshot to the rest of the network.
+4. Reconciliation - the process by which snapshots received from other nodes is reflected on to bring the network state
+   knowledge up-to-date.
+
+Finally, the specification expands on cases where conflict might occur and how a node should resolve it.
+
```
gammazero, line 121: Should this specification propose ways to offer indexer incentives? Pay for querys, pay for results, pay for advertisements ingested, etc.?

IPNI_FEDERATION.md
masih: It does also have link to previous. I'll update the description here 👍

IPNI_FEDERATION.md
masih: Though useful, this feature seems sufficiently complicated to need its own specification?

Aside from that, I think what's documented in the current spec doesn't technically limit that behaviour. For example, there is nothing that would stop a group of collaborating indexers to agree on only indexing advertisements up to the depth 1000 within the last week. The federation protocol described here would still work as long as all the participating nodes only include providers with such criteria. 

WDYT @gammazero ?

gammazero: I suppose that is the purpose of this spec - to make indexing consistent between indexers that cover identical index content.  It means that if indexer-1 covers providers A and B, indexer-2 covers B and C, and indexer-3 covers C and A, these cannot be in the same federation even though they can check each other's consistency. Or, are they each part of 2 federations?

I can see index allocation, and incentivization as completely different specs.

IPNI_FEDERATION.md
gammazero: Answered above - No, that should be called out in a seperate spec as long as it is not precluded by this one.

IPNI_FEDERATION.md
masih: I think there certainly is but checking indexers' behaviour is an independent problem: it applies to a single index too, not just a cluster of federated ones.

Adding that to the federation protocol itself seems like possible future extension. But I am not sure if it is a blocker for having a federation.

IPNI_FEDERATION.md
masih: I think incentives is also a separate problem than "Can I find what I am looking for no matter which indexer I search?"

Those seem like time consuming factors to figure out which will most likely be application/service specific. I am not sure if we can meaningfully pin them down in a spec.

Another thing I would keep in mind is a wider ecosystem incentives for CIDs and content addressed data specifically. For example, how Boost works in terms of deal making or retrieval will impact incentives of indexing.

Overall, I too think it is something to certainly revisit. I would probably start with a concrete use case like what would those numbers be for cid.contact first before adding them to a spec.

IPNI_FEDERATION.md
gammazero: This seems unfair to the faster indexer, because all the other indexers will query the mirror of the fastest. Maybe we need a strategy that strongly encourages other indexers to add content from others' mirrors to their own.  

For example, an indexer cannot pull advertisement data directly from another mirror, but instead must sync some portion of the mirror and then get the advertisements from the local mirror after syncing.

The fast source indexer may then redirect others, for a short time, to the destination indexer for the synced content. Or, maybe there is some check that if one indexer copies advertisement data from another indexer's mirror, the indexer that received the data subsequently serves it from its mirror.

IPNI_FEDERATION.md
```diff
@@ -0,0 +1,408 @@
...
+**Resolution Strategy:**  
+This represents a divergence in the advertisement chain, and careful reconciliation is required:
+
+1. Instance A and B determine which advertisement is the later one in the chain of advertisements provided by P1.
+2. They converge on the advertisement has been published the latest, i.e. has the shorter distance to P1's current
+   advertisement head.
+
+While traversing advertisement chains, instances should prefer pulling ad chain from eachother's mirror instead of going
+directly to the provider where possible.
```
gammazero, line 230: This seems unfair to the faster indexer, because all the other indexers will pull from the mirror of the fastest. Maybe we need a strategy that strongly encourages other indexers to add content from others' mirrors to their own.

For example, an indexer cannot pull advertisement data directly from another mirror, but instead must sync some portion of the mirror and then get the advertisements from the local mirror after syncing. This forces the receiving indexer to have a copy of the data in its own mirror.

The original source indexer may then redirect others, for a short time, to the destination indexer for the synced content. Or, maybe there is some check that if one indexer copies advertisement data from another indexer's mirror, the indexer that received the data subsequently serves it from its mirror.

IPNI_FEDERATION.md
masih: > I suppose that is the purpose of this spec

It all boils down to the words "Collaborating IPNI indexers".
The current spec aggregates provider lists across all indexers.  This means one indeed cannot have an indexer that only indexes a proportion of providers instead of all them. 

In the case you described, at least 2 of those indexers need to agree on the heuristic for selecting providers. Right now that heuristic in cid.contact is simply "All that you can get your hands on but not the ones you have not seen for a week".  If indexer-1, 2 and 3 follow the current heuristic and are part of the same federation, then the should all end up with A, B and C in their providers list.

The specification of what that heuristic is, IMO is beyond the scope of this spec. I think we can both imagine case where it can grow quite complicated, and could potentially lead into a lot of head scratching for users trying to look things up.

Considering the spec does not prohibit the federation of a set of indexers that agree upon the same heuristic to select providers, my vote would be to keep the scope small and aim for a federation of cid.contact-like indexers going first.

In that world, one could use DNS names to form separate federations across indexers that only care about "hot off the wire" records. For example, today.cid.contact could be a new indexer part of a separate federation of indexers that only index ads published within the last 24 hours.

IPNI_FEDERATION.md
masih: In think in practice, the fastest indexer would/should have rate limiting in place, and other indexers should scatter traffic across multiple sources simply to pull ads from different providers in parallel.

The key point I am hoping to get across with this sentence is to say: try going to mirrors first before going to providers. That doesn't mean the same mirror all the time. 

The "fastest indexer" would also depend on where the providers and indexers physically are. I expect that the indexer network would naturally grow around the organisational structure of its providers. For example, if there is a heavy presence of providers in Australia , it is likely that there would be more than one indexer in Australia to cope with the demand. At this point I find myself heavily speculating which is probably a hint that we might be over thinking it.

Another factor here is the loosely defined concept of Mirror itself: there is no specification yet that documents the API for a mirror. We are in the process of figuring out its economics, rate limits etc. 

So i think we would probably be OK revisiting this once there is more data.

Thoughts?

gammazero: Is there anything that encourages one indexer to help another, when those indexers are not operated by the same organization? I guess just that it allows the other indexer to supply the ad data and reduces load on the first.

In other blockchain indexing ecosystems, the indexers are interested in handling as many queries as they can for the data they index - so they do not want to help other indexers. In the IPNI ecosystem, it appears that indexers are interested in doing as little of the required work as possible - so want other indexers to handle as many queries as possible.

gammazero: For now, I am OK with allowing indexers to share their mirror as much or as little as possible, with whatever limits they see fit, and specifying this a some later time.

masih: Captured this https://github.com/ipni/specs/issues/29

gammazero (approved):

ischasny (approved):

IPNI_FEDERATION.md
```diff
@@ -0,0 +1,425 @@
...
+
+The snapshot mechanism also offers a strategic advantage. In the event of any system failures or communication
+breakdowns, snapshots can be used to quickly restore or reconcile the state of an instance, minimizing data loss and
+ensuring rapid recovery.
+
+A snapshot consists of:
+
+* **Epoch**: a monotonically increasing value that represents the time at which snapshot was formed.
+* **Vector Clock**: the monotonically increasing vector clock value that corresponds to the IPNI instance.
+* **Provider to Ingest State Map**: a map of provider ID to the latest advertisement CID processed by the IPNI
```
aschmahmann, line 148: Does "processed" mean there's an expectation here that if a client see "height 10" that all of the advertisements up to height 10 will be queryable (by clients or other IPNI nodes), or does that mean they've simply heard about the latest CID/height. Similarly, must height have been validated?

Note: I get that you may not want to be prescriptive here around caching / loadbalancing strategies (e.g. could a client end up with inconsistent state due to a load balancer routing the data to different backends rather than being sticky), either way it'd probably be helpful to call this out for any client implementers.

IPNI_FEDERATION.md
```diff
@@ -0,0 +1,425 @@
...
+  instance, along with an optional advertisement height if known by the indexer. The height is provided as a _hint_ to
+  other indexers in order to aid the reconciliation and conflict resolution process.
+
+As the IPNI federation protocol continues to evolve, these specification components—initialization and periodic
+snapshots—act as the backbone, ensuring that the system remains resilient, efficient, and, most importantly,
+consistently synchronized across all instances. Future enhancements may introduce additional layers of optimization or
+fault-tolerance, but the essence will remain rooted in these foundational principles.
+
+### Snapshot Exchange
```
aschmahmann, line 157: As @hsanjuan mentioned elsewhere you might consider if a more prescriptive syncing scheme (e.g. having IPNI nodes post CRDT updates to a shared kv-store like ipfs-cluster uses https://github.com/ipfs/go-ds-crdt) would help you here.

This could reduce the amount of data emitted by snapshots as well as help with the synchronization among many nodes rather than leaving the "how to find the latest state" work up to the individual instances.

aschmahmann (commented):
Great to see the interest here 🙏. Apologies for the late review and hope it's of use.

hsanjuan:
Notes from going over:

* ` instances can use metadata or summary vectors of their snapshots. These summaries allow instances to quickly ascertain if they need the full snapshot or if they're already up-to-date` : I don't see how the state is represented or how "summaries" are represented.  I don't think you can have a syncing mechanism that involves exchanging full snapshots. That doesn't scale with huge indexes nor indexes that are updated very frequently by several parties. There needs to be a notion of "delta", and how the state is represented is important to have a way to create those deltas. ie. perhaps representing the state as a merkle-dag can facilitate diffing and snapshot sharing: https://ieeexplore.ieee.org/document/9049566 (a perhaps better alternative to merkle-crdts from go-ds-crdt mentioned above).
* Broadcast mechanisms are not detailed: how does a broadcast here work? Careless broadcast will become the first bottleneck on a federated system. Is that an HTTP POST endpoint? Is that other protocol?
* How are snapshots authenticated? Is the CID a hash of the full snapshot? What happens when snapshots are huge?

IPNI_FEDERATION.md
gammazero: Yes, "processed" means that the advertised content should be queryable, otherwise that status has little useful external meaning. At very least it should mean that the indexer has pulled the index content from the provider and the corresponding advertisements are available from the indexer's mirror. I prefer it to mean the content is also queryable.

IPNI_FEDERATION.md
```diff
@@ -0,0 +1,425 @@
...
+  Whether nodes are structured in a mesh, star, or hierarchical format, the protocol remains agnostic.
+* **Strong consistency**: While the IPNI federation protocol aims to ensure data consistency and availability, it does
+  not aim to atomically replicate every piece of indexed data across all nodes. For example, it is possible for some
+  records to remain partially available on one node until deletions are propagate through the network.
+
+## Terminology
+
+* **Epoch**: An epoch in the context of the IPNI federation protocol refers to a specific time period or iteration of
+  the protocol's operation. It can be thought of as a version or generation number that increments each time there's a
+  major operation or update, aiding in synchronization efforts across instances.
```
gammazero, line 92: Who determines what an epoch consists of? Different IPNI instances may have different rates of progress for different index providers, depending on network location, bandwidth limits, etc. If any single IPNI instance tries to declare an epoch based on its view of indexing progress, then other IPNI instance may never be able to reach consistency because they will always be ahead for some index providers.

I think the means an epoch is a set of agreed-on progress points for all index providers in the federation, that each IPNI node can independently determine. A distance from the root of each provider's advertisement chain could serve this purpose. A date may not be a good example unless the IPNI nodes have access to a shared clock.

willscott: this sounds like consensus on a vector clock?

IPNI_FEDERATION.md
```diff
@@ -0,0 +1,425 @@
...
+
+* **Reconciliation**: Reconciliation is the process of ensuring that two or more IPNI instances have consistent index
+  records. It involves comparing the content of instances, identifying discrepancies, and making updates so that all
+  participating instances have the same content.
```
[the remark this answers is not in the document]
masih: > Reconciliation requires IPNI nodes to wait for each other to reach the same point

Not necessarily. The ingest velocity is a separate thing from what the lookup returns in response to a given CID.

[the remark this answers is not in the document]
willscott: > but not progress beyond that point, until the pre-arranged point for all other index providers is reached.

is it possible for new entries to be added for a cid, but with a timestamp, such that responses to queries at an earlier epoch are then filtered back out, and as such the response is consistent for that epoch even when the node has moved past it?

IPNI_FEDERATION.md
gammazero: Different epochs can contain different sets of index providers, but an epoch has a specific set of index providers associated with it.

gammazero (changes requested):
Need to resolve some of the questions before merging this.

IPNI_FEDERATION.md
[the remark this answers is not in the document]
gammazero: I am thinking about the case where something in a later epoch deletes items from an earlier epoch, making those items no longer queryable. Would that prevent reaching consensus at that earlier epoch?

PR #35, opened 2025-02-09
Specify multihash sampling APIs

masih:
Add a specification that outlines a multihash sampling specification for IPNI, also known as reverse index.

[Rendered](https://github.com/ipni/specs/blob/masih/mh-sampling/IPNI_MH_SAMPLING.md)

### DIFF
--- a/IPNI_MH_SAMPLING.md
+++ b/IPNI_MH_SAMPLING.md
@@ -101,7 +101,7 @@ The key rationales for the **IPNI Multihash Sampling** API are:
 
 ### Determinism and Consistency
 
-The **IPNI Multihash Sampling** uses deterministic algorithms with a uniform seed to produce consistent, repeatable
+The **IPNI Multihash Sampling** uses deterministic algorithms with a uniform beacon to produce consistent, repeatable
 sampling sequences. This reduces sampling bias and ensures the same results for repeated queries. Key components
 include:
 
@@ -112,21 +112,33 @@ include:
 * **Federation Epoch Awareness**: The API maintains consistent multihash sampling across indexer instances within the
   same federation by considering the federation epoch.
 
-* **Seed-Based Sampling**: Users can specify a seed derived from decentralized randomness beacons like Drand. This
-  guarantees deterministic sampling and just-in-time candidate selection. Queries repeated with the same seed produce
+* **Seed-Based Sampling**: Users can specify a beacon derived from decentralized randomness beacons like Drand. This
+  guarantees deterministic sampling and just-in-time candidate selection. Queries repeated with the same beacon produce
   identical results, facilitating reproducibility.
 
-The sampling process utilizes the PCG (Permuted Congruential Generator) random number generator, seeded to ensure
-deterministic output. To efficiently perform sampling, two pre-emptive data organizational strategies are proposed:
+The sampling process employs the PCG (Permuted Congruential Generator) random number generator, using a seed derived
+from a beacon to ensure deterministic output. The 128-bit PCG seed is extracted from a 32-byte beacon through the
+following steps:
+
+- The 32-byte beacon is split into two 16-byte segments.
+- The least significant bytes of each 16-byte segment are used to form the high and low 64-bit seed values,
+  respectively.
+
+If the provided beacon is smaller than 32 bytes, it is padded with zeros to reach the required seed size. The use of a
+32-byte beacon, even though half might be discarded, is driven by the convenience of aligning with existing distributed
+randomness beacons like DRAND, which often provide larger outputs, and allows for future extensibility by reserving
+additional space for potential enhancements or increased complexity in random seed generation.
+
+To efficiently perform sampling, two pre-emptive data organizational strategies are proposed:
 
 1. **Counting Multihashes per Context ID**: Indexers pre-calculate the total multihash count per provider ID for each
    context ID.
 2. **Sorting Multihashes by Hash Value**: Indexers store multihash records grouped by provider ID and context ID, sorted
    by hash value. This complements a global multihash index linked to provider information.
 
-The sampling process involves initializing the PCG with the seed and selecting the required indices with the upper bound
-being the total multihashes per context ID. The selected multihashes are then retrieved from the sorted list by hash
-value.
+The sampling process involves initializing the PCG with the beacon and selecting the required indices with the upper
+bound being the total multihashes per context ID. The selected multihashes are then retrieved from the sorted list by
+hash value.
 
 While other deterministic sampling algorithms don't require these pre-emptive data organizational strategies, the
 proposed method is more efficient for large datasets. It also supports efficient bulk deletions and deeper meta-analysis
@@ -150,11 +162,14 @@ Samples a set of multihashes ingested by an IPNI indexer for a given provider ID
 - **Path Parameters**:
     - `provider-id`: (string, required) The multibase encoded peer ID of the provider.
         - Example: `12D3KooWN34sqTaMfZE3ReELyVF3no3qU7883Mi6j2VWsv6dwhPL`
-    - `context-id`: (string, required) The multibase encoded context ID.
+    - `context-id`: (string, required) The base64 encoded context ID.
+        - Example: `AXESIFVcxmAvWdc3BbQUKlYcp2Z2DuO2w5Fo4jmIC8IbMX00`
 
 - **Query Parameters**:
-    - `seed`: (optional) The seed for deterministic sampling. Ensures repeatability of samples.
-    - `max`: (optional) The maximum number of multihashes to return. Defaults to one if unspecified.
+    - `beacon`: (string, optional) The hex encoded randomness beacon for deterministic sampling. Ensures repeatability
+      of samples. Must not exceed 32 bytes.
+    - `max`: (integer, optional) The maximum number of multihashes to return. Defaults to one if unspecified. Must be
+      greater than zero, with a maximum of 10.
     - `federation_epoch`: (optional) The IPNI federation epoch, currently only accepting zero, pending review of IPNI
       federation protocol.
 
--- a/IPNI_MH_SAMPLING.md
+++ b/IPNI_MH_SAMPLING.md
@@ -1,0 +1,274 @@
+# IPNI Multihash Sampling
+
+![wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+
+**Author(s)**:
+
+- [Masih Derkani](https://github.com/masih)
+
+**Maintainer(s)**:
+
+- [Masih Derkani](https://github.com/masih)
+- [Will Scott](https://github.com/willscott)
+
+* * *
+
+**Abstract**
+
+This document outlines the specification for the **IPNI Multihash Sampling**, a complimentary component of the IPNI
+ecosystem designed to facilitate efficient and scalable sampling of multihash records. The API provides a standardized
+interface for querying multihash records advertised by a provider for a given context ID, enabling use cases such as
+retrieval checking, data integrity verification, and decentralized reputation systems.
+
+## Table of Contents
+
+- [Introduction](#introduction)
+- [Motivation](#motivation)
+- [Background](#background)
+- [Specification](#specification)
+    - [Rationale](#rationale)
+    - [Determinism and Consistency](#determinism-and-consistency)
+    - [API](#api)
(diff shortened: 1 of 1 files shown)


willscott (approved):

IPNI_MH_SAMPLING.md
```diff
@@ -0,0 +1,274 @@
...
+
+Samples a set of multihashes ingested by an IPNI indexer for a given provider ID and Context ID.
+
+- **Path Parameters**:
+    - `provider-id`: (string, required) The multibase encoded peer ID of the provider.
+        - Example: `12D3KooWN34sqTaMfZE3ReELyVF3no3qU7883Mi6j2VWsv6dwhPL`
+    - `context-id`: (string, required) The multibase encoded context ID.
+
+- **Query Parameters**:
+    - `seed`: (optional) The seed for deterministic sampling. Ensures repeatability of samples.
```
bajtos, line 156: In my limited experience with RNGs, the seed is usually a number. 

However, the DRAND beacon is a string, e.g. `3439d92d58e47d342131d446a3abe264396dd264717897af30525c98408c834f`. I believe this string is a hex-encoded representation of some binary data (a hash digest 32-bytes long?) so we can easily convert it into a big number - 23622477310538551577113686196064208635296596430983311976502637696987544158208. Would your implementation support such big numbers?

As a developer integrating with this new endpoint, I don't know what kind of seed values the API expects and supports.

Let's make the spec a bit more detailed, please, and describe the seed value type plus any constraints like maximum numeric value or maximum string length.

bajtos (commented):
Great spec, @masih!

The proposal looks good to me. I have a few suggestions for improving the API documentation; see below.

IPNI_MH_SAMPLING.md
```diff
@@ -0,0 +1,274 @@
...
+### API
+
+#### `GET /ipni/v0/sample/{provider-id}/{context-id}`
+
+Samples a set of multihashes ingested by an IPNI indexer for a given provider ID and Context ID.
+
+- **Path Parameters**:
+    - `provider-id`: (string, required) The multibase encoded peer ID of the provider.
+        - Example: `12D3KooWN34sqTaMfZE3ReELyVF3no3qU7883Mi6j2VWsv6dwhPL`
+    - `context-id`: (string, required) The multibase encoded context ID.
```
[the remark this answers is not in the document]
bajtos: In my FRC, I reverse-engineered the ContextID constructed by Curio from PieceCID and PieceSize, see https://github.com/filecoin-project/FIPs/pull/1089/files#diff-6e350230c51dc5f5f28b9435a04a777ecced6f2f425f5febd5aedd4cdcb59a97R186-R225

AFAICT, they convert the binary ContextID value to a string using the hex encoding with no multibase prefix. Example ContextID:

```
1B0000000800000000D82A5828000181E203922020FC66377F35B3780A65368D0DA3FE1862EFF9FB36DB1D416CE7B94C2E1062E80E
```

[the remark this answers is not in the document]
bajtos: Based on the above, I believe ContextID can be any string. Maybe such context ID values are not following the spec, but they seem to be used in practice and supported by the current IPNI implementation running at https://cid.contact.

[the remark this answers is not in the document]
masih: The context ID as far as IPNI is concerned is opaque bytes. It is also encoded as such in the JSON response you kindly pointed to as an example. Here is a small go test that decodes it (and that Context ID so happens to be a CID):

```go
func TestContextIDDecoding(t *testing.T) {
	bytes, err := base64.StdEncoding.DecodeString("AXESIFVcxmAvWdc3BbQUKlYcp2Z2DuO2w5Fo4jmIC8IbMX00")
	require.NoError(t, err)
	length, cid, err := cid.CidFromBytes(bytes)
	require.NoError(t, err)
	t.Log(cid)
	t.Log(length)
	// outputs:
	// bafyreicvltdgal2z243qlnaufjlbzj3goyhohnwdsfuoeomibpbbwml5gq
	// 36
}
```

I think it's probably easier to stick to the current format of it in the URL parameters, which is *base64 encoded byte*. I will adjust the spec and thanks for pointing this out.

IPNI_MH_SAMPLING.md
masih: For the purposes of spec and ease of use we can: accept string representation of DRAND beacon directly. More specifically: maximum 64 characters, hex-encoded 32 bytes. I will also rename seed to beacon to avoid confusion with the rng seed in a general sense.

Diving deep into implementation: 
The RNG i intend to use is PCG, which is seed with 128-bits of internal state, typically supplied as 2 64-bit high/low integers.

I can convert the 32byte long DRAND beacon into two 64-bit unsigned integers using the least significant 64 bits of the supplied big integer.

masih: I have adjusted the spec to simply use "beacon". I have also expanded the section on determinism to concretely specify how the seed for PCG random number generator is derived from the 32 byte beacon.

IPNI_MH_SAMPLING.md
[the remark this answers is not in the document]
masih: Marking as resolved. Please let me if there is anything that I might have missed.

IPNI_MH_SAMPLING.md
```diff
@@ -0,0 +1,289 @@
...
+### API
+
+#### `GET /ipni/v0/sample/{provider-id}/{context-id}`
+
+Samples a set of multihashes ingested by an IPNI indexer for a given provider ID and Context ID.
+
+- **Path Parameters**:
+    - `provider-id`: (string, required) The multibase encoded peer ID of the provider.
+        - Example: `12D3KooWN34sqTaMfZE3ReELyVF3no3qU7883Mi6j2VWsv6dwhPL`
+    - `context-id`: (string, required) The base64 encoded context ID.
```
bajtos, line 165: Continuing the discussion from https://github.com/ipni/specs/pull/35#discussion_r1952255385

> The context ID as far as IPNI is concerned is opaque bytes. It is also encoded as such in the JSON response you kindly pointed to as an example. Here is a small go test that decodes it (and that Context ID so happens to be a CID):
> 
> ```go
> func TestContextIDDecoding(t *testing.T) {
> 	bytes, err := base64.StdEncoding.DecodeString("AXESIFVcxmAvWdc3BbQUKlYcp2Z2DuO2w5Fo4jmIC8IbMX00")
> 	require.NoError(t, err)
> 	length, cid, err := cid.CidFromBytes(bytes)
> 	require.NoError(t, err)
> 	t.Log(cid)
> 	t.Log(length)
> 	// outputs:
> 	// bafyreicvltdgal2z243qlnaufjlbzj3goyhohnwdsfuoeomibpbbwml5gq
> 	// 36
> }
> ```
> 
> I think it's probably easier to stick to the current format of it in the URL parameters, which is _base64 encoded byte_. I will adjust the spec and thanks for pointing this out.

I agree with you to use the same format as IPNI returns in `GET /cid/{cid}` response, i.e. _base64-encoded bytes_.

However, this does not match what I described in my FRC based on the understanding I had at the time. I'd like to double check that the advertisements provided by SPs in JSON over HTTP contain ContextID values in base64 encoding too.

I think that should be true, since Curio depends on libipni to convert the advertisement structure to JSON string, and Go's JSON marshaller converts `byte[]` values to base64-encoded strings.

I'd like to double check this against some real advertisements to confirm.

masih: Here is one: https://cid.contact/cid/QmfSxgcqGMj1CicQiFR7UJNfaL1XHvPj7uzpDex9g59Q3v

bajtos: Thank you for the example!

Here is what I found:

https://cid.contact/providers/12D3KooWJ91c6xQshrNe7QAXPFAaeRrHWq2UrgXGPf8UmMZMwyZ5

```json5
{
  "AddrInfo": {
    "ID": "12D3KooWJ91c6xQshrNe7QAXPFAaeRrHWq2UrgXGPf8UmMZMwyZ5",
    "Addrs": [
      "/dns/f03303347-market.duckdns.org/https/http-path/%2Fipni-provider%2F12D3KooWJ91c6xQshrNe7QAXPFAaeRrHWq2UrgXGPf8UmMZMwyZ5"
    ]
  },
  "LastAdvertisement": {
    "/": "baguqeerakqnjugtecgj5hhfmnh46pthk5k3vy6vf4bw3vpmwmudknlatun5a"
  },
  // ...
}
```

https://f03303347-market.duckdns.org/ipni-provider/12D3KooWJ91c6xQshrNe7QAXPFAaeRrHWq2UrgXGPf8UmMZMwyZ5/ipni/v1/ad/baguqeerakqnjugtecgj5hhfmnh46pthk5k3vy6vf4bw3vpmwmudknlatun5a

```json5
{
  "Addresses": [
    "/dns/f03303347-market.duckdns.org/https/http-path/%2Fipni-provider%2F12D3KooWJ91c6xQshrNe7QAXPFAaeRrHWq2UrgXGPf8UmMZMwyZ5"
  ],
  "ContextID": {
    "/": {
      "bytes": "ghsAAAAIAAAAANgqWCgAAYHiA5IgIFeksuf2VqNvrrRUxrvA+itvJhrDRju06ThagW6ULKw2"
    }
  },
  "Entries": {
    "/": "bafyreictdikh363qfxsmjp63i6kup6aukjqfpd4r6wbhbiz2ctuji4bofm"
  },
  // ...
}
```

The ContextID value `ghsAAAAIAAAAANgqWCgAAYHiA5IgIFeksuf2VqNvrrRUxrvA+itvJhrDRju06ThagW6ULKw2` represents `PieceInfo: {Size:34359738368 PieceCID:baga6ea4seaqfpjfs473fni3pv22fjrv3yd5cw3zgdlbumo5u5e4fvalosqwkynq}
`; see https://go.dev/play/p/bXjgXRD7bLH

All is good 👍🏻

IPNI_MH_SAMPLING.md
```diff
@@ -0,0 +1,290 @@
...
+following steps:
+
+- The 32-byte beacon is split into two 16-byte segments.
+- The least significant bytes of each 16-byte segment are used to form the high and low 64-bit seed values,
+  respectively.
+
+If the provided beacon is smaller than 32 bytes, it is padded with zeros to reach the required seed size. The use of a
+32-byte beacon, even though half might be discarded, is driven by the convenience of aligning with existing distributed
+randomness beacons like DRAND, which often provide larger outputs, and allows for future extensibility by reserving
+additional space for potential enhancements or increased complexity in random seed generation.
```
bajtos, line 130: It makes me wonder if it's a good idea to implement the extraction on the IPNI side. I am concerned this will make it difficult for clients using a different randomness than DRAND. For example, if my source of randomness produces 128-bit values, and I want IPNI to use this value, I would have to "upscale" my 128-bit value to a 32-byte beacon using an algorithm that reverses the steps that IPNI will perform to "downscale" the beacon to 128 bits.

What if we update the specification to require the beacon (or seed) to be exactly 128-bits long, i.e. 32 characters when using hex encoding, and let the clients to decide how to downscale DRAND's 256-bit beacon to 128-bit seed. (Maybe we can keep the algorithm you described above as a recommendation.)

I am sharing this as an idea to consider. From Spark's perspective, your current spec based on the DRAND beacon works great for us.

bajtos (approved):
👍🏻

IPNI_MH_SAMPLING.md
masih: >  I would have to "upscale" my 128-bit value to a 32-byte beacon 

You would not. Note that the spec states "Must not exceed 32 bytes." A client is free to supply a smaller hex encoded seed.

> What if we update the specification to require the beacon (or seed) to be exactly 128-bits long, i.e. 32 characters when using hex encoding, and let the clients to decide how to downscale DRAND's 256-bit beacon to 128-bit seed.

I have expressed the rationale for allowing a larger seed (that so happens to match DRAND's 32 byte long beacon): for future extensibility. If there is a need, it would be harder to go to a larger value than smaller one. That combined with the fact that the spec accepts smaller seeds would offer a reasonable tradeoff for both DRAND and non-drand users.

Having said that, I have no hard held opinions on this one.

masih: Reworded the seed generation to be clearer and simplified the approach slightly.

bajtos (approved):
:shipit:

